PR #8, opened 2024-12-19
Optimize the QpMap processing and add a compute filter to the encoder.

zlatinski:
Optimize the QpMap to use inlined command buffer transfer operations or direct handling of the images.
Add a compute filter as a preprocessor for the encoder to do last row and last column replication, color conversion, and scalling.


zlatinski:
Reviewed and verified by Raju Konda.

PR #9, opened 2024-12-21
Compute 16 bit and image row and column replication support

zlatinski:
Add 16-bit compute filter support

Add support for row and column replication. Two modes:
1: (default) replicate the last row and column to the padding area;
2: replicate only one row and one column to the padding area;


zlatinski:
I'm merging this one since there are no objections so far and there are follow-up changes that we need to push in depending on these commits.

PR #12, opened 2025-01-13
ProgramConfig: add av1 decode operation

dabrain34:
As h26x, the av1 codec should be able to be passed to the cli.


dabrain34:
cc @zlatinski

zlatinski (approved):
Looks good. Thank you for the fix!

PR #14, opened 2025-01-16
Use vulkan sdk for the vulkan header files and shaderc and tools

zlatinski:
Use vulkan sdk for the vulkan header files and shaderc and tools instead of using locally built tools and headers.
Updated build instructions.
Remove the shaderc and tools scripts and vulkan headers.
Patched the cmake files to use the Vulkan SDK.


dabrain34:
You need to update the github actions `.github/worflows/main.yml` to remove the use of `update_external_sources.sh` and `update_external_sources.bat`

dabrain34:
one problem I see with this change is that CTS needs to build 32 bits and AFAIK there is no 32 bits version of the vulkan SDK. Issue open on Lunarg issue tracker

dabrain34:
https://vulkan.lunarg.com/issue/view/678a2ac629714d508d09d3f5

zlatinski:
Thank you, [dabrain34](https://github.com/dabrain34),

> one problem I see with this change is that CTS needs to build 32 bits and AFAIK there is no 32 bits version of the vulkan SDK. Issue open on Lunarg issue tracker

For CTS, we'll use the CTS-built compiler and elementary streams only for testing.
We don't want to manage local compiler builds for this project anymore. It is  the source of a lot of build problems.

zlatinski:
This change was already merged with the decoder as a library

PR #16, opened 2025-01-29
Only show warning when testing in verbose mode

lordalcol:
The warning was printed many times while testing with the Vulkan CTS. Since --testOutOfOrderRecording is being passed, it's obvious that the intention is to use it "for testing", therefore the warning is fairly intrusive.


dabrain34:
LGTM thanks!

PR #18, opened 2025-03-07
Add Duck IVF parsing support to the elementary stream.

vasn1k:
Add support for .ivf files when loading AV1 compressed video files and using elementary streams.

### DIFF
--- a/vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
+++ b/vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
@@ -18,6 +18,10 @@
 #include <fstream>
 #include "mio/mio.hpp"
 #include "VkDecoderUtils/VideoStreamDemuxer.h"
+#define DKIF_FRAME_CONTAINER_HEADER_SIZE 12
+#define DKIF_HEADER_MAGIC *((const uint32_t*)"DKIF")
+#define DKIF_FILE_HEADER_SIZE 32
+#define USE_SIMPLE_MALLOC 1
 
 class ElementaryStream : public VideoStreamDemuxer {
 
@@ -32,11 +36,38 @@ public:
         , m_height(defaultHeight)
         , m_bitDepth(defaultBitDepth)
         , m_videoCodecType(forceParserType)
+#ifndef USE_SIMPLE_MALLOC
         , m_inputVideoStreamMmap()
+#endif
         , m_pBitstreamData(nullptr)
         , m_bitstreamDataSize(0)
         , m_bytesRead(0) {
 
+#ifdef USE_SIMPLE_MALLOC
+        FILE* handle = fopen(pFilePath, "rb");
+        if (handle == nullptr) {
+            printf("Failed to open video file %s\n", pFilePath);
+            m_bitstreamDataSize = 0;
+            m_pBitstreamData = nullptr;
+            return;
+        }
+
+        fseek(handle, 0, SEEK_END);
+        size_t size = ftell(handle);
+        uint8_t* data = (uint8_t*)malloc(size);
+
+        if (data == nullptr) {
+            printf("Failed to allocate memory for video file: %i\n", (uint32_t)size);
+            m_bitstreamDataSize = 0;
+            return;
+        }
+
+        m_bitstreamDataSize = size;
+        fseek(handle, 0, SEEK_SET);
+        fread(data, 1, size, handle);
+        m_pBitstreamData = data;
+        fclose(handle);
+#else
         std::error_code error;
         m_inputVideoStreamMmap.map(pFilePath, 0, mio::map_entire_file, error);
         if (error) {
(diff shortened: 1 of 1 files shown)


zlatinski (approved):
LGTM Thank you!

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -18,6 +18,10 @@
 #include <fstream>
 #include "mio/mio.hpp"
 #include "VkDecoderUtils/VideoStreamDemuxer.h"
+#define DKIF_FRAME_CONTAINER_HEADER_SIZE 12
+#define DKIF_HEADER_MAGIC *((const uint32_t*)"DKIF")
+#define DKIF_FILE_HEADER_SIZE 32
+#define USE_SIMPLE_MALLOC 1
```
dabrain34, line 24: I'm  a bit confused about this USE_SIMPLE_MALLOC. Can you add a comment to explain the use of that ?

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -18,6 +18,10 @@
 #include <fstream>
 #include "mio/mio.hpp"
 #include "VkDecoderUtils/VideoStreamDemuxer.h"
+#define DKIF_FRAME_CONTAINER_HEADER_SIZE 12
+#define DKIF_HEADER_MAGIC *((const uint32_t*)"DKIF")
+#define DKIF_FILE_HEADER_SIZE 32
```
dabrain34, line 23: would be nice to use structure and get some information from the header such as the codec.
You can get inspiration from https://github.com/Igalia/ESExtractor/blob/main/lib/eseivfstream.cpp

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -32,11 +36,38 @@ class ElementaryStream : public VideoStreamDemuxer {
...
+#endif
         , m_pBitstreamData(nullptr)
         , m_bitstreamDataSize(0)
         , m_bytesRead(0) {
 
+#ifdef USE_SIMPLE_MALLOC
+        FILE* handle = fopen(pFilePath, "rb");
+        if (handle == nullptr) {
+            printf("Failed to open video file %s\n", pFilePath);
+            m_bitstreamDataSize = 0;
```
dabrain34, line 50: no need for that as its already set to null and zero above

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -32,11 +36,38 @@ class ElementaryStream : public VideoStreamDemuxer {
...
+            m_pBitstreamData = nullptr;
+            return;
+        }
+
+        fseek(handle, 0, SEEK_END);
+        size_t size = ftell(handle);
+        uint8_t* data = (uint8_t*)malloc(size);
```
dabrain34, line 57: I would avoid this approach for very large file. I would read the size of the frame by the frame header and move the offset to this size.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -32,11 +36,38 @@ class ElementaryStream : public VideoStreamDemuxer {
...
+
+        if (data == nullptr) {
+            printf("Failed to allocate memory for video file: %i\n", (uint32_t)size);
+            m_bitstreamDataSize = 0;
```
dabrain34, line 61: no need  for this code as it hasnt been modified so far

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
```diff
@@ -83,7 +130,11 @@ class ElementaryStream : public VideoStreamDemuxer {
     }
 
     virtual ~ElementaryStream() {
+#ifdef USE_SIMPLE_MALLOC
```
dabrain34, line 133: maybe use #ifndef

dabrain34 (changes requested):

zlatinski:
We should remove the USE_SIMPLE_MALLOC from this change. @vasn1k, per our offline discussion, please do that before we merge this change.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
vasn1k: I'll remove it the define. It's there for debugging purposes.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
vasn1k: The stream doesn't currently have a way to relay back this information, so there is no current need to extract it and hold it.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
vasn1k: This shouldn't have been empty.

vk_video_decoder/libs/VkDecoderUtils/ElementaryStream.cpp
vasn1k: USE_SIMPLE_MALLOC was supposed to be disabled. The code is kept here to prevent merge conflicts later.

dabrain34:
@vasn1k could you make the change needed for this PR ?

PR #20, opened 2025-03-13
VulkanPerDrawContext::RecordCommandBuffer: use member vars

### LINKED ISSUE #19: RecordCommandBuffer in VulkanVideoUtils shadows VulkanPerDrawContext class members
The method `RecordCommandBuffer` takes as argument `descriptorSetLayoutBinding` and `samplerYcbcrConversion` which are also member of the same class which shadow the use of the variable, see

https://github.com/KhronosGroup/Vulkan-Video-Samples/blob/44fa4da5c781d3dfdb4c9eef4a593ce17d4f95a0/common/libs/VkCodecUtils/VulkanVideoUtils.h#L655

The use of correct variable must be clarified.

dabrain34:
RecordCommandBuffer should use its own member values instead of passing it through its API.

Avoid vars shadowing detected by CTS

Fixes #19


zlatinski (approved):
Thanks for the fix, @dabrain34! LGTM.

PR #23, opened 2025-03-14
Add missing copy of input data from file to memory for YUV444P

dabrain34:
Also add support frame count calculations for YUV444P and YUV422P

Backport from https://github.com/nvpro-samples/vk_video_samples/pull/129


dabrain34:
cc @krajunv

zlatinski (approved):
Thanks for picking up this change from nvpro samples. LGTM.

PR #25, opened 2025-03-18
DecoderConfig: y4m should be off by default

dabrain34:
As there is a flag to enable y4m in the decoder config, the default value should be false and enabled by the flag on demand.


zlatinski (approved):
LGTM. Thank you!

PR #29, opened 2025-04-01
DecoderConfig: enable stream demuxing

dabrain34:
By default, if FFmpeg is present, the stream
demuxer should be enabled.


dabrain34:
All the media decode failed if we used `--codec codec` except H264 annexB content.

zlatinski (approved):
LGTM. Thank you!

PR #5, opened 2024-12-16
Logger: add a log system instead of std::cout or printf

dabrain34:
## Description

Add a logger with level from ERROR to DEBUG

There are 2 different to call logs:

 - LOG_S_DEBUG which is a log with level 4 and acts as std::cout
 -  LOG_DEBUG which is a log with level 4 and acts as printf("fmt\n", args);
 -  LOG_DEBUG_CAT which is a log with level 4 with a category as string and acts as printf("fmt\n", args);

There are 4 kind of logs:
  - LOG_DEBUG
  - LOG_INFO
  - LOG_WARN
  - LOG_ERROR

## Type of change

features

### Intel(R) Iris(R) Xe Graphics (ADL GT2) (8086:46a6) / Intel open-source Mesa driver Mesa 26.3.0-devel (git-bef5b14778) / Ubuntu 24.04.5 LTS

Total Tests:    85
Passed:         58
Crashed:         0
Failed:          0
Not Supported:  24
Skipped:         3 (in skip list)
Success Rate: 100.0%

PR #30, opened 2025-04-03
log: disable shaders dump

dabrain34:
Disable the computer shader log as it is displayed by CTS AV1 encode tests


dabrain34:
it would be nice to review and use the logger implemented in #5

PR #34, opened 2025-04-23
Fix Intel ANV support

dabrain34:
Multiple fixes were necessary to address the support of Intel Mesa ANV driver:

 - the image create format needs to be set properly
 - ANV driver has a separate transfer queue from the decode queue. Add support of this queue.
 - ANV driver needs more back buffers then NVIDIA driver. Set it according to the caps.


dabrain34:
related to https://github.com/nvpro-samples/vk_video_samples/issues/71

dabrain34:
@lolzballs can you have a look to this change too ?

lolzballs (approved):
Did not test with RADV, but the changes make sense and it works on AMD's Windows driver.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -1261,6 +1261,47 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
...
+            &waitDstStageMask,             // const VkPipelineStageFlags*  pWaitDstStageMask;
+            1u,                            // uint32_t                                             commandBufferCount;
+            &m_transferCommandBuffers[0],           // const VkCommandBuffer*               pCommandBuffers;
+            1u,                            // uint32_t                                             signalSemaphoreCount;
+            &videoDecodeCompleteSemaphore,                   // const VkSemaphore*                   pSignalSemaphores;
+        };
+        assert(VK_NOT_READY == m_vkDevCtx->GetFenceStatus(*m_vkDevCtx, transferCompleteFence));
+        VkResult result = m_vkDevCtx->MultiThreadedQueueSubmit(VulkanDeviceContext::TRANSFER, m_vkDevCtx->GetTransferQueueFamilyIdx(),
+                                                               1, &transferSubmitInfo, transferCompleteFence);
+        result = m_vkDevCtx->WaitForFences(*m_vkDevCtx, 1, &transferCompleteFence, true, gFenceTimeout);
```
zlatinski, line 1299: Stalling the pipeline?

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -1261,6 +1261,47 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
         }
     }
 
+    if (m_useTransferOperation == VK_TRUE && m_transferCommandPool != VK_NULL_HANDLE) {
```
zlatinski, line 1264: Really? We are creating and destroying resources mid-frames?

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
```diff
@@ -1118,7 +1118,7 @@ int VkVideoDecoder::DecodePictureWithParameters(VkParserPerFrameDecodeParameters
     VkVideoEndCodingInfoKHR decodeEndInfo = { VK_STRUCTURE_TYPE_VIDEO_END_CODING_INFO_KHR };
     m_vkDevCtx->CmdEndVideoCodingKHR(frameDataSlot.commandBuffer, &decodeEndInfo);
 
-    if (m_useTransferOperation == VK_TRUE) {
+    if (m_useTransferOperation == VK_TRUE && m_transferCommandPool == VK_NULL_HANDLE) {
```
zlatinski, line 1121: Have you tried using the filter?

--enablePostProcessFilter 0

This should work on any implementations.

zlatinski (changes requested):
I don't believe this is the most effective way to implement the transfer. Consider using a filter for your transfer. Furthermore, we can derive a filter class that uses transfer-only operations. The filter framework doesn't allocate and free resources on each frame and uses a semaphore for synchronization (it doesn't stall the pipeline).

common/libs/VkShell/Shell.cpp
```diff
@@ -201,6 +207,15 @@ void Shell::CreateSwapchain() {
     m_ctx.swapchain = VK_NULL_HANDLE;
     m_ctx.extent.width = (uint32_t)-1;
     m_ctx.extent.height = (uint32_t)-1;
+
+
+    if (image_count <= caps.minImageCount) {
+        image_count = caps.minImageCount + 1;
```
zlatinski, line 213: We are adding +1 here and then also in Shell::CreateBackBuffers() Do we need to add 1 to both palces?

Otherwise, this change looks good to me.

dabrain34: I double checked and indeed nvidia/AMD(Win)/Intel(Win) driver support to create back buffers with `caps.minImageCount + 1` when mesa driver needs at least +2 to work properly, so I guess there is an issue in the mesa driver. I will keep +2 with a comment to explain

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
dabrain34: outdated as transferQueue usage has been removed.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
lolzballs: Why would we prefer the compute based copy rather than the transfer based which uses `vkCmdCopyImage`?

If an implementation has dedicated transfer HW it would be more efficient to use the transfer queue.

dabrain34: So vkCmdCopyImage with dedicated transfer queue will be more efficient that the compute based copy ?

lolzballs: Yes since it might use DMA based copy, so that the compute units are free to do other stuff or even be off the conserve power.

dabrain34: so I should keep the transfer queue usage seen that it is not used by nvidia

common/libs/VkCodecUtils/VulkanFrame.cpp
```diff
@@ -186,7 +186,7 @@ int VulkanFrame<FrameDataType>::AttachSwapchain(const Shell& sh)
     const uint32_t queueFamilyIndices = m_videoRenderer->m_vkDevCtx->GetGfxQueueFamilyIdx();
     imageCreateInfo.pQueueFamilyIndices = &queueFamilyIndices;
     imageCreateInfo.initialLayout = VK_IMAGE_LAYOUT_PREINITIALIZED;
-    imageCreateInfo.flags = 0;
+    imageCreateInfo.flags = VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
```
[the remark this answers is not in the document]
zzoon: > Why is this required? From the spec: If image was created with a [multi-planar format](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#formats-multiplanar), and the image view’s aspectMask is one of VK_IMAGE_ASPECT_PLANE_0_BIT, VK_IMAGE_ASPECT_PLANE_1_BIT or VK_IMAGE_ASPECT_PLANE_2_BIT, the view’s aspect mask is considered to be equivalent to VK_IMAGE_ASPECT_COLOR_BIT when used as a framebuffer attachment.
> 

Right, I think this is just a workaround.. but still the spec also says:

```
If the image has a multi-planar format, subresourceRange.aspectMask is VK_IMAGE_ASPECT_COLOR_BIT,
and usage includes VK_IMAGE_USAGE_SAMPLED_BIT, then the format must be identical to
the image format and  the sampler to be used with the image view must enable 
sampler Y′CBCR conversion.
```

So, I think, the IV format should be identical to the image format.  Thoughts?

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
dabrain34: I brought back the use of this transfer queue as it does not harm if the decode and transfer are on the same queue.

dabrain34: I compared the use of transfer queue again Process Filter and the result is quite eleoquent:

 - `--noPresent --postProcessFilterType 1` : Frame 301, FPS: 1197.6
 -  `--noPresent --postProcessFilterType 0` : Frame 301, FPS: 2461.32

common/libs/VkCodecUtils/VulkanFrame.cpp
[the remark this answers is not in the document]
zlatinski: Yes, so this above sentence is for the usage with graphics pipeline. It says: YCbCr images must be used with a sampler in the graphics/compute pipelines or as individual planes.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.cpp
zlatinski: > Why would we prefer the compute based copy rather than the transfer based which uses vkCmdCopyImage?
> If an implementation has dedicated transfer HW it would be more efficient to use the transfer queue.
I'm not saying that compute would be more efficient here on all HW. But if we are going to allocate and free resources on each frame and stall the pipeline with fences, then the compute filter would be much more efficient.

The filter interface is a generic class. Instead of using the compute implementation, one can inherit a transfer-based filter. This class provides pre-allocated command buffers, fences, and semaphores. So,

When you are allocating the object for the filter, just create an instance of class transfer, not compute. The rest of the code would work the same. The filter is using a semaphore to synchronize with the video queues without stalling the pipeline.

common/libs/VkShell/Shell.cpp
zlatinski: OK, thanks for explaining this!

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -68,7 +68,7 @@ struct DecoderConfig {
         gpuIndex = -1;
         forceParserType = VK_VIDEO_CODEC_OPERATION_NONE_KHR;
         decoderQueueSize = 5;
-        enablePostProcessFilter = -1,
+        postProcessFilterType = 0, //FilterType::DISABLED
```
zlatinski, line 71: Post-processing filter 0 means don't do any conversion of the pixels. This can be used for linear-to-optimal conversion. We've used these parameters for a while now. Please don't break this interface for us.

zlatinski:
I suggest this MR be split into multiple topic-specific MRs. Some of the changes are ready to be merged; others need more work.

common/libs/VkShell/Shell.cpp
dabrain34: I can create a separate MR for this change

dabrain34: https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/40

common/libs/VkCodecUtils/VulkanFrame.cpp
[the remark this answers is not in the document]
zzoon: Yeah, I mean in the code, the IV format is NOT identical to the image format, which I think it's not valid. So we need to fix this?

common/libs/VkCodecUtils/DecoderConfig.h
dabrain34: Sorry about this API breakage. I have a created a separate MR #41 to handle this change. Dropping the commit.

common/libs/VkCodecUtils/VulkanFrame.cpp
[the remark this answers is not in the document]
zzoon: This violates the spec for sure. 

VVL complains like that:
```
vkCreateImageView(): pCreateInfo->format VK_FORMAT_R8_UNORM is different from VkImage 0x190000000019 format (VK_FORMAT_G8_B8R8_2PLANE_420_UNORM). Formats MUST be IDENTICAL unless VK_IMAGE_CREATE_MUTABLE_FORMAT BIT was set on image creation.
The Vulkan spec states: If image was not created with the VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT flag, or if the format of the image is a multi-planar format and if subresourceRange.aspectMask is VK_IMAGE_ASPECT_COLOR_BIT, format must be identical to the format used to create image (https://docs.vulkan.org/spec/latest/chapters/resources.html#VUID-VkImageViewCreateInfo-image-01762)

```

So we need to set VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT or do not try to create iv for each plane.
Thoughts?

dabrain34:
#40 has been created to separate the MR
#43 has been created to separate the MR

PR #40, opened 2025-05-09
Shell: create backBuffer after SwapChain

dabrain34:
Create enough back buffers
according to the caps of the device.

ANV needs more back buffers than NVIDIA driver.


zlatinski (approved):
LGTM. Thank you, Stephane!

PR #42, opened 2025-05-15
Integrated the parser with CTS

zzoon:
Currently, CTS is using a customed parser in VVS for decoding but using an AV1 encoder here in VVS, which creates unnecessary complexity. We agreed to unify the usage of VVS and it should be this VVS here.

To support this, we need to do the following:

- Include "vkDefs.hpp" if CTS.
- Build the static lib of the parser correctly, which will be used in CTS.
- Add one param `isAnnexB` to the VkParserInitDecodeParameters, which is required in CTS.


dabrain34:
LGTM

PR #43, opened 2025-05-19
VulkanImage: set the image create format

dabrain34:
ANV driver needs the flag VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT

Separate PR following the discussion of https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/34#discussion_r2093136540

### DIFF
--- a/common/libs/VkCodecUtils/VulkanFrame.cpp
+++ b/common/libs/VkCodecUtils/VulkanFrame.cpp
@@ -186,7 +186,7 @@ int VulkanFrame<FrameDataType>::AttachSwapchain(const Shell& sh)
     const uint32_t queueFamilyIndices = m_videoRenderer->m_vkDevCtx->GetGfxQueueFamilyIdx();
     imageCreateInfo.pQueueFamilyIndices = &queueFamilyIndices;
     imageCreateInfo.initialLayout = VK_IMAGE_LAYOUT_PREINITIALIZED;
-    imageCreateInfo.flags = 0;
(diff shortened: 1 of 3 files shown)


dabrain34:
@lolzballs @zzoon

common/libs/VkCodecUtils/VulkanFrame.cpp
```diff
@@ -186,7 +186,7 @@ int VulkanFrame<FrameDataType>::AttachSwapchain(const Shell& sh)
     const uint32_t queueFamilyIndices = m_videoRenderer->m_vkDevCtx->GetGfxQueueFamilyIdx();
     imageCreateInfo.pQueueFamilyIndices = &queueFamilyIndices;
     imageCreateInfo.initialLayout = VK_IMAGE_LAYOUT_PREINITIALIZED;
-    imageCreateInfo.flags = 0;
+    imageCreateInfo.flags = VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
```
lolzballs, line 189: As discussed in the original thread, `MUTABLE` is required on this specific image creation because `VkImageResourceView::Create()` always creates per-plane image views.

I think this is the easiest path to be spec compliant here, so the changes here are fine to me. One alternative would be to only create the per-plane image-views if they are really needed. For example, in the case of this this testFrameImage, it seems like only the `COLOR` aspect is used.

zzoon: Agree

zlatinski (approved):
I'm fine with this change. Thank you!

Issue #4, opened 2024-12-11
Create a first release of the project to be used by CTS


dabrain34:
I would like to propose a first revision to start with v0.1.0. What do you think @zlatinski ?

zlatinski:
OK, sure, Stephan, sounds good.

dabrain34:
Fixed with https://github.com/KhronosGroup/Vulkan-Video-Samples/releases/tag/release-v0.2.0 which will be used by CTS in a next release.

PR #45, opened 2025-05-21
Version: add a first version number for VVS

dabrain34:
Use 0.1.0 as a startup version.

### DIFF
--- a/common/include/Version.h
+++ b/common/include/Version.h
@@ -1,0 +1,22 @@
+#ifndef VVS_VERSION_H
+#define VVS_VERSION_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define VVS_VERSION_MAJOR 0
+#define VVS_VERSION_MINOR 1
+#define VVS_VERSION_PATCH 0
+
+#define STRINGIFY_HELPER(x) #x
+#define STRINGIFY(x) STRINGIFY_HELPER(x)
+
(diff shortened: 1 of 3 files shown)


dabrain34:
For now I havent thought about a use case where the decoder and the encoder could have a dedicated version as we can use this version number to tell what decoder and encoder was available in the package.

Any suggestion @lolzballs @zlatinski ?

dabrain34:
WIP for #4

ceyusa:
most of the projects I have checked, use a template that generates the version.h in configuration time, using, must of the time the latest git tag to get the version of the source code. Otherwise maintainers would have to keep in sync two version holders: git and this header.

common/include/VkVSVersion.h
```diff
@@ -0,0 +1,22 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define VKVS_VERSION_MAJOR 0
+#define VKVS_VERSION_MINOR 1
+#define VKVS_VERSION_PATCH 0
+
+#define STRINGIFY_HELPER(x) #x
+#define STRINGIFY(x) STRINGIFY_HELPER(x)
```
ceyusa, line 13: these lines are kind of useless, fmpov

ceyusa: dah... I forgot that

> If you want to stringize the result of expansion of a macro argument, you have to use two levels of macros.

PR #52, opened 2025-06-03
CMake: remove non working and unused CMakefiles

dabrain34:
Since ab3b3b3, the build of decoder and encoder have been merged into the root CMake file. No need for this additional unused files.


srinathkr-nv:
Thank you @dabrain34; LGTM.

PR #56, opened 2025-06-06
VulkanDeviceContext: initialize transfer queue

dabrain34:
As the transfer queue can be different from the decode queue, this queue must be detected and associated to a valid number.

Fix typo in the name of the m_transferQueue member var.


zzoon:
Looks good to me. 
I'm going to drop the 1st commit in the #60

PR #65, opened 2025-06-19
decoder: fix unused variable warning

dabrain34:
Declare result only once and use the value
to decide to continue to initialize the image set.


zlatinski (approved):
LGTM. Thank you!

Issue #54, opened 2025-06-04
Mesa: Encode: H26x encode fails with division by zero


dabrain34:
```
~/DEV/SCRIPTS/run-radv.sh ./vk_video_encoder/demos/vk-video-enc-test -c h265 -i ~/DEV/VK-GL-CTS/BUILD/external/vulkancts/modules/vulkan/vulkan/video/yuv/176x144_30_i420.yuv -o /tmp/result.h265 --inputWidth 176 --inputHeight 144
```

```
Thread 1 "vk-video-enc-te" received signal SIGFPE, Arithmetic exception.
0x000055555556f7dd in VkVideoGopStructure::GetPositionInGOP (this=0x555555690e50, gopState=..., gopPos=..., firstFrame=false, framesLeft=29)
    at /DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoGopStructure.h:176
176	        gopPos.inGop = (uint8_t)(gopState.positionInInputOrder % m_gopFrameCount);
(gdb) bt
#0  0x000055555556f7dd in VkVideoGopStructure::GetPositionInGOP (this=0x555555690e50, gopState=..., gopPos=..., firstFrame=false, framesLeft=29)
    at /DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoGopStructure.h:176
#1  0x00005555555800a6 in VkVideoEncoderH265::EncodeFrame (this=0x5555557639c0, encodeFrameInfo=...)
    at/DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderH265.cpp:385
#2  0x000055555559e154 in VkVideoEncoder::EncodeFrameCommon (this=0x5555557639c0, encodeFrameInfo=...)
    at /DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp:333
#3  0x000055555559e945 in VkVideoEncoder::StageInputFrame (this=0x5555557639c0, encodeFrameInfo=...)
    at /DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp:433
#4  0x000055555559dae0 in VkVideoEncoder::LoadNextFrame (this=0x5555557639c0, encodeFrameInfo=...)
[...]

PR #72, opened 2025-06-20
EncoderConfig: set default preferredGopFrameCount

dabrain34:
As some implementations (mesa) do report preferredGopFrameCount to 0 and seen that the minimum value is always chosen, use UINT8_MAX as a default value if its zero.

Fixing a divide by zero issue in VkVideoGopStructure.h+176

Fixing #54


zlatinski (approved):
LGTM. Thank you for the fix!

dabrain34:
Working on the mesa driver [implementation](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35767), David Roscoe highlights an important part of the [spec](https://registry.khronos.org/vulkan/specs/latest/man/html/VkVideoEncodeH264RateControlInfoKHR.html) saying: 

`If it is 0, the rate control algorithm may assume an implementation-dependent GOP length / IDR period. And this is basically what happens, so if anything I'd say our preferred values would be 0.` So its allowed to have zero in preferredGOP`

So I guess the approach to set the GOPFrameCount to Max if preferred value is 0, is the right approach.

Issue #73, opened 2025-06-23
Build issue on aarch64 architecture


elima:
Building CTS on aarch64 is failing on NvVideoParser:

```
aarch64-unknown-linux-gnu-g++: error: unrecognized command-line option '-mavx512f'
aarch64-unknown-linux-gnu-g++: error: unrecognized command-line option '-mavx512bw'
```

Seems like AVX512 compiler flags are being added to non-x86 archs by mistake.

srinathkr-nv:
> Building CTS on aarch64

This suggests that CTS is being built on an aarch64 CPU but #74 seems to involve fixes related to cross-compilation of CTS. Is #74 handling both cases i.e. native compilation on aarch64 and compiling-for-aarch64-on-x86_64 ?

dabrain34:
Lets create a ci job in another MR to check cross target

PR #74, opened 2025-06-23
Improves cross-compilation detection and fix compiler flags in NvVideoParser

elima:
AVX512 flags are currently being added regardless of the arch. This patch fixes that and improves cross-compilation handling.

Fixes https://github.com/KhronosGroup/Vulkan-Video-Samples/issues/73


zlatinski (approved):
LGTM. Thanks for the fix!

dabrain34 (changes requested):
it would be nice to have a cross compilation target in github actions, linux aarch64 to validate the patch.

PR #81, opened 2025-07-08
fix shadowing variable

dabrain34:
cmake: add -Wshadow and fix related issues
  
 - Remove the use of extra VkResult var declaration.
 -  VulkanDeviceContext: use a different deviceUuid var name

Got warnings building CTS with 0.3.0 branch release


srinathkr-nv:
Thank you for these fixes, @dabrain34. LGTM.

PR #82, opened 2025-07-08
cmake: create FIND_VULKAN_HEADERS macro

dabrain34:
This macro allows to detect only the headers
and will be used by CTS.
It fixes a bug with the CTS buildbot which does not
have pkg-config and fails to build the
vulkan-loader which is not necessary in the case
of CTS build.


dabrain34:
I have been reconsidering the vulkan headers detection and up to my amalysis, we want to use the vulkan sdk to avoid downloading the vulkan-headers repo but in the case the vulkan sdk version is not sufficient, we should download the vulkan-headers, so I'm rewriting the logic in a next commit.

dabrain34:
The PR should be ready to review now,

cmake/FindVulkanSDK.cmake
```diff
@@ -63,30 +65,48 @@ MACRO(FIND_VULKAN_SDK minimum_major_version minimum_minor_version minimum_patch_
...
+            GIT_TAG main
+        )
+        FetchContent_MakeAvailable(vulkan-headers)
+
+        # Set Vulkan headers path (we are using the downloaded headers)
+        set(VULKAN_HEADERS_INCLUDE_DIR ${vulkan-headers_SOURCE_DIR}/include CACHE PATH "Path to Vulkan include headers directory" FORCE)
+        message(STATUS "VULKAN_HEADERS_INCLUDE_DIR: ${VULKAN_HEADERS_INCLUDE_DIR}")
+    else()
+        # System version is good, now call find_package to create targets
+        find_package(Vulkan REQUIRED)
```
zzoon, line 93: I think we need to handle the case where it found via pkg-config here.?   
For example, in my case, the above pkg-config search found a vulkan(which is the latest), but here the library is set as `/usr/lib/...`

zzoon:
Also, not included in this PR though, the below thing in FindVulkanSDK.cmake doesn't detect vulkan 1.4

```
# Extract major and minor version
        file(STRINGS "${VULKAN_HEADERS_INCLUDE_DIR}/vulkan/vulkan_core.h" VK_VERSION_1_3_LINE
            REGEX "^#define VK_API_VERSION_1_3")
        if(VK_VERSION_1_3_LINE)
            set(VK_MAJOR_VERSION 1)
            set(VK_MINOR_VERSION 3)
        endif()
 
```

dabrain34:
> Also, not included in this PR though, the below thing in FindVulkanSDK.cmake doesn't detect vulkan 1.4
> 
> ```
> # Extract major and minor version
>         file(STRINGS "${VULKAN_HEADERS_INCLUDE_DIR}/vulkan/vulkan_core.h" VK_VERSION_1_3_LINE
>             REGEX "^#define VK_API_VERSION_1_3")
>         if(VK_VERSION_1_3_LINE)
>             set(VK_MAJOR_VERSION 1)
>             set(VK_MINOR_VERSION 3)
>         endif()
>  
> ```

should be fixed now

cmake/FindVulkanSDK.cmake
dabrain34: @srinathkr-nv lets keep the logic then. Can you review this patch ? I would use this version to create a 0.3.1 release.

srinathkr-nv:
The one minor thing I would prefer is to place all changes related to Vulkan 1.4 (like the update to `VK_MINOR_VERSION`, or changing the string matching logic to look for the Vulkan 1.4 line) in a single commit.

Currently, the above updates are in the "cmake: check for Vulkan SDK before downloading headers" commit whereas the actual version bump is in "cmake: update SDK minimum version".

Otherwise, these changes look good to me.

srinathkr-nv:
Thank you @dabrain34; the current set of commits LGTM.

PR #83, opened 2025-07-11
encoder: fix type conversion from double to uint32_t

dabrain34:
The following conversion might lead to precision loss, use a double instead of uin32_t to keep the precision.


srinathkr-nv:
LGTM.

PR #87, opened 2025-08-12
vulkan include cleanup

dabrain34:
Centralize the include of vulkan headers in `vulkan_interfaces.h` which then will be used by CTS to build VVS source files.

This work allows CTS to build VVS source files using vkDefs.h when DE_BUILD_VVS is set which expose the vulkan symbols generated in vkVulkan_c.inl.

This will avoid to have 2 different includes version for the vulkan headers.


common/include/vulkan_interfaces.h
```diff
@@ -23,6 +23,8 @@
 #ifdef DE_BUILD_VIDEO
 #include "vkDefs.hpp"
 using namespace vk;
+#elif defined(DE_BUILD_VVS)
+#include "vkDefs.h"
```
ceyusa, line 27: AFAIK there's no need to add this since it's already defined just above, we just have to add the 

`add_compile_definition(DE_BUILD_VIDEO)` in CTS external/vulkan_video_samples

zzoon:
@dabrain34  do you have any branch for CTS, related to this change?

ceyusa:
> @dabrain34 do you have any branch for CTS, related to this change?

imo it should be an update of the patch I posted, since the headers won't need to be generated now, just use those that vulkancts already use.

zzoon:
for the third patch, it doesn't look to affect VVS itself. So we can land it unless anyone else is against it.

dabrain34:
@zlatinski @srinathkr-nv any comment on this change ?

common/include/vulkan_interfaces.h
```diff
@@ -19,10 +19,21 @@
...
 
-// This flag is defined in CTS.
+// This flag is defined in CTS to build the parser source files which are also in the vk namespace in external/vulkancts/modules/vulkan/video
 #ifdef DE_BUILD_VIDEO
 #include "vkDefs.hpp"
 using namespace vk;
+// This flag is defined in CTS to build VVS library and provide the vulkan definitions from CTS in vkVulkan_c.inl
+#elif defined(VVS_CTS_VULKAN_HEADER)
+#include "vkDefs.h"
+#if defined(VK_USE_PLATFORM_WIN32_KHR)
```
ceyusa, line 29: Overall I'm OK with the changes.

Stupid question: what happens if you just do

``` 
#undef VK_USE_PLATFORM_WIN32_KHR
``` 

??

As a personal comment about coding taste: I'm not a big fan of juggling with symbols in code. I prefer to decouple the presentation layer with the video processing, so the cmake scripts would decide which files to include in the compilation.

dabrain34: Thanks for your concern and I agree with your position about the extra defines used. So I came back to my initial solution which was not working because of the mssing windows.h include. So now VK_USE_PLATFORM_WIN32_KHR is used only for Vulkan purpose. Otherwise I changed it with _WIN32 guard which is more relevant. Now VK_USE_PLATFORM_WIN32_KHR wont be defined in CTS.

PR #94, opened 2025-08-22
Allow setting Vulkan headers include path in cmake

lordalcol:
Also add an option to fail the build instead of downloading sources. This allows the CTS to re-use the Vulkan headers already fetched by other components, which also means aligning with the same Vulkan headers version.

With the new option FAIL_IF_NO_HEADERS, the CTS can enforce re-using its own headers, and if they are not present of their version is not compatible with the one required by vulkan-video-samples, it will fail the build instead of downloading new ones.


lordalcol:
Not needed anymore as it has been resolved with CTS CL17896

PR #95, opened 2025-08-28
EncoderConfigH264: use ADAPTIVE_TRANSFORM_AUTOSELECT

dabrain34:
In order to use the baseline or main profile,
transform_8x8_mode_flag should not be set to true. Use the auto select which is more appropriate here.

In case ADAPTIVE_TRANSFORM_ENABLE is used, disable it and warn user that this option can not be used with baseline and main profile.


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -312,6 +312,11 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
 
     if (adaptiveTransformMode == ADAPTIVE_TRANSFORM_ENABLE) {
         pps->flags.transform_8x8_mode_flag = true;
+         if ((profileIdc == STD_VIDEO_H264_PROFILE_IDC_BASELINE) ||
```
srinathkr-nv, line 315: The indentation / alignment of these new lines seems a bit off, but the profile check makes sense to me. Once the indentation / alignment issue is resolved, this change is good to be merged.

PR #98, opened 2025-09-04
DecoderConfig: show help if no input specified

dabrain34:
Print an error and exit if the input filename
has not been specified.
Print an error if the video file is not valid and exit.


srinathkr-nv:
This change looks fine to me.

dabrain34:
I made a small change in the logic, can you have another look @srinathkr-nv ?

srinathkr-nv:
The updated logic also LGTM.

PR #102, opened 2025-09-25
common: remove unused variable

dabrain34:
Remove m_ycbcrPrimariesConstants which is
unused in the codebase.


dabrain34:
Hello @zlatinski I opened this MR to discuss the use m_ycbcrPrimariesConstants which is unused in VulkanFilterYuvCompute.cpp.

It allows to avoid a warning on Windows building this with CTS.

Can you give your input on what you prefer to do with this unused variable ?

PR #104, opened 2025-09-25
FindVulkanSDK: fix reconfigure

dabrain34:
In case of missing Vulkan SDK, VVS
checkouts and builds its own version
of the VulkanSDK but it might fail on
the second configure.


zzoon:
LGTM here.

PR #105, opened 2025-09-26
EncoderConfig: warn about production feature only in verbose

dabrain34:
--testIntraRefreshMidway and --testSkipIntraRefreshStart should be reported as "used for testing only" in verbose mode.

The warning was printed many times while testing with the Vulkan CTS. Since arguments are being passed, it's obvious that the intention is to use it "for testing", therefore the warning is fairly intrusive.


dabrain34:
@srinathkr-nv ^

PR #110, opened 2025-10-24
VkEncoderConfig{H264,H265}: Check if --slices is beyond max

lolzballs:
Without this check, the sample app will happily try to fill out more slice info structures than supported, eventually writing past the bounds of those arrays and causing segfault.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -15,6 +15,7 @@
 */
 
 #include "VkVideoEncoder/VkEncoderConfigH264.h"
+#include "VkVideoEncoder/VkVideoEncoderH264.h"
 
 int EncoderConfigH264::DoParseArguments(int argc, char* argv[])
 {
@@ -24,6 +25,10 @@ int EncoderConfigH264::DoParseArguments(int argc, char* argv[])
         if (args[i] == "--slices") {
             if (++i >= argc || sscanf(args[i].c_str(), "%u", &sliceCount) != 1) {
                 fprintf(stderr, "invalid parameter for %s\n", args[i - 1].c_str());
+                return -1;
+            }
+            if (sliceCount > VkVideoEncoderH264::MAX_NUM_SLICES) {
+                fprintf(stderr, "this application supports a max of %d slices\n", VkVideoEncoderH264::MAX_NUM_SLICES);
                 return -1;
             }
         } else if (args[i] == "--profile") {
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
@@ -16,6 +16,7 @@
 
 #include <math.h>       /* sqrt */
 #include "VkVideoEncoder/VkEncoderConfigH265.h"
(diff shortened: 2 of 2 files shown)


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -77,6 +78,11 @@ int EncoderConfigH265::DoParseArguments(int argc, char* argv[])
                 fprintf(stderr, "invalid parameter for %s\n", args[i - 1].c_str());
                 return -1;
             }
+            if (sliceCount > VkVideoEncoderH265::MAX_NUM_SLICES) {
+                fprintf(stderr, "this application supports a max of %d slices\n", sliceCount);
```
dabrain34: It should be 
```suggestion
                fprintf(stderr, "this application supports a max of %d slices\n", VkVideoEncoderH265::MAX_NUM_SLICES);
```

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -26,6 +27,10 @@ int EncoderConfigH264::DoParseArguments(int argc, char* argv[])
                 fprintf(stderr, "invalid parameter for %s\n", args[i - 1].c_str());
                 return -1;
             }
+            if (sliceCount > VkVideoEncoderH264::MAX_NUM_SLICES_H264) {
+                fprintf(stderr, "this application supports a max of %d slices\n", sliceCount);
```
dabrain34:
```suggestion
                fprintf(stderr, "this application supports a max of %d slices\n", VkVideoEncoderH264::MAX_NUM_SLICES_H264);
```

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -77,6 +78,11 @@ int EncoderConfigH265::DoParseArguments(int argc, char* argv[])
...
+                return -1;
+            }
+        } else if (args[i] == "--profile") {
```
srinathkr-nv: This is a duplicated condition, and I suspect that it'll break the functionality to specify the codec profile.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderH265.h
```diff
@@ -24,6 +24,7 @@
 
 class VkVideoEncoderH265 : public VkVideoEncoder {
 
+public:
```
srinathkr-nv: Instead of moving the `public` qualifier up, it might be better to move the two enum definitions down, closer to the constructor. This is to avoid making the `VkVideoEncodeFrameInfoH265` definition accessible outside the class (which I don't see a need for in this change).

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderH264.h
```diff
@@ -25,8 +25,11 @@ class VkVideoEncoderH264 : public VkVideoEncoder {
 
 public:
 
-    enum { NON_VCL_BITSTREAM_OFFSET = 4096 };
     enum { MAX_NUM_SLICES_H264 = 64 };
```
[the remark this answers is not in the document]
lolzballs: Agree, no need for the suffix as it's already namespaced.

srinathkr-nv:
LGTM too.

PR #111, opened 2025-10-25
ci: add arm 32 bits job

dabrain34:


dabrain34:
Including #109 to fix the arm 32 bits build of VVS

PR #106, opened 2025-10-06
tests: introduce testing framework

dabrain34:
This framework provides a testing infrastructure for Vulkan
Video encode and decode operations.

Key features:
- Multi-codec support: H.264, H.265, AV1, VP9 (decode); H.264, H.265,
AV1 (encode)
- Automatic sample download with checksum verification (MD5/SHA256)
- Decoder validation with MD5 (fluster approach)
- Encoder validation via decoder roundtrip testing
- External test suite support (Fluster/Soothe formats)
- Flexible filtering by codec, pattern, and skipped status
- Platform-agnostic (Windows/Linux)

Usage:
- See tests/README.md for details
 
The application use standard exit code for most
of the app exit such as EXIT_FAILURE or EXIT_SUCCESS.

Add custom VVS_EXIT_UNSUPPORTED
when the queue does not support the family requested.

The application version is now defined in VkVSCommon.h

PR #119, opened 2025-11-26
VulkanAV1Decoder: Clear segmentation flags when not enabled

lolzballs:
## Description

When segmentation is not enabled, none of the segmentation related fields should be set. The pStd structure can be reused between pictures, and in some cases, the flags would persist between frames.

## Type of change

Bug fix.

## Issue (optional)

This fixes [decode_av1_superres_8bit](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/106/files#diff-aa52a7008fa79a1d9c3cf7340a53754575df5041ee468ac4b11857510b10f9bdR177) on AMD implementations.

### AMD Radeon RX 7900 XTX (RADV NAVI31) / radv Mesa 25.3.5-arch1.1 / Arch Linux

Total Tests:    70
Passed:         44
Crashed:         0
Failed:          0
Not Supported:   1
Skipped:        25 (in skip list) -- **Improvement from 26 on main**
Success Rate: 100.0%

### DIFF
--- a/tests/skipped_samples.json
+++ b/tests/skipped_samples.json
@@ -297,34 +297,6 @@
       "date_added": "2026-01-19"
     },
     {
-      "name": "av1_superres_8bit",
-      "format": "vvs",
-      "drivers": [
-        "radv"
-      ],
-      "platforms": [
-        "all"
-      ],
-      "reproduction": "always",
-      "reason": "md5 mismatch",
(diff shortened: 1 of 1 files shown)


lolzballs:
There shouldn't be, could you try the latest main?

Also what MD5 do you get?

PR #128, opened 2025-12-12
FindShaderc: refactor download logic into macro and use shallow clones

dabrain34:
#  FindShaderc: refactor download logic into macro and use shallow clones
    
    Create download_shaderc_and_dependencies macro to consolidate shaderc
    and dependency download logic.

    Rework the shaderc detection especially for system one.
    
    Add GIT_SHALLOW TRUE to all FetchContent
    declarations for faster downloads.

  #  cmake: link VkVideoUtils to vulkan loader
    
    Add target_link_libraries for VkVideoUtils to link against the vulkan
    loader. This fixes build failures when building without system vulkan
    packages installed, allowing the FetchContent-built vulkan-loader to
    be used.

 # docs: add Fedora packages and libvulkan-dev to build requirements
    
    Add Fedora/DNF package installation instructions alongside existing
    Ubuntu/Debian instructions. Also add libvulkan-dev to Ubuntu packages
    for vulkan loader development files.

 #   FindVulkanSDK: use shallow clone for FetchContent
    
    Add GIT_SHALLOW TRUE to vulkan-headers and vulkan-loader FetchContent
    declarations to speed up downloads by only fetching the tagged commit
    without full git history.

### DIFF
--- a/cmake/FindShaderc.cmake
+++ b/cmake/FindShaderc.cmake
@@ -1,60 +1,16 @@
 # Now check if we need to build shaderc and its dependencies
-set(USE_SYSTEM_SHADERC ON CACHE BOOL "Use system installed shaderc" FORCE)
 option(USE_SYSTEM_SHADERC "Use system installed shaderc" ON)
-set(USE_SHADERC ON CACHE BOOL "Find shaderc either from the system or build it" FORCE)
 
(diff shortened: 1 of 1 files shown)


zlatinski (approved):
Thank you for the clean-up and the improvements, Stephane!

PR #130, opened 2025-12-12
Get count_trailing_zeros() to compile for MSVC 32-bit

pdaniell-nv:
This was causing Vulkan CTS build to fail on MSVC Windows builds for 32-bit:
> P:\git\gerrit.khronos.org\vk-gl-cts\external\vulkan-video-samples\src\vk_video_decoder\libs\NvVideoParser\include\cpudetect.h(20,41): error C3861: '_     ': identifier not found
> (compiling source file '../../../external/vulkan-video-samples/src/vk_video_decoder/libs/NvVideoParser/src/NextStartCodeC.cpp')

This showed up with vulkan-cts-1.4.5.0, but probably has been around before that.

### DIFF
--- a/vk_video_decoder/libs/NvVideoParser/include/cpudetect.h
+++ b/vk_video_decoder/libs/NvVideoParser/include/cpudetect.h
@@ -13,11 +13,19 @@ enum SIMD_ISA
 
 static int inline count_trailing_zeros(unsigned long long resmask)
 {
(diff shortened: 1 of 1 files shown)


zlatinski (approved):
LGTM. Thank you, Piers for the fix!

zlatinski (approved):

Issue #134, opened 2025-12-18
Warnings when building on Linux with clang


rg3igalia:
There are a few warnings when building VVS with clang on Linux, which have in the past appeared when VVS is pulled and built as part of CTS. Most of them can be fixed with the following changes:

```diff
diff --git a/common/libs/VkCodecUtils/DecoderConfig.h b/common/libs/VkCodecUtils/DecoderConfig.h
index b194889..deb108f 100644
--- a/common/libs/VkCodecUtils/DecoderConfig.h
+++ b/common/libs/VkCodecUtils/DecoderConfig.h
@@ -21,7 +21,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <iostream>
-#include <fstream>
 #include <string>
 #include <vector>
 #include <functional>
@@ -114,7 +113,7 @@ struct DecoderConfig {
                     return rtn;
                 }},
             {"--version", "-v", 0, "Show the version",
-                [argv](const char **, const ProgramArgs &a) {
+                [](const char **, const ProgramArgs &a) {
                     fprintf(stdout, "%s\n", VKVS_VERSION_STRING);
                     exit(EXIT_SUCCESS);
                     return true;
diff --git a/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h b/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
index ffe3a06..e2aefd9 100644
--- a/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
+++ b/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
@@ -19,7 +19,6 @@
 
 #include <sstream>
 #include <vulkan_interfaces.h>
-#include "VkCodecUtils/Helpers.h"
 #include "VkCodecUtils/VulkanDeviceContext.h"
 #include "VkCodecUtils/VkBufferResource.h"
 #include "VkCodecUtils/VulkanSamplerYcbcrConversion.h"
@@ -271,13 +270,13 @@ public:
             switch (pDescriptorWrite->descriptorType) {
             case VK_DESCRIPTOR_TYPE_SAMPLER:
                 descriptorSize = m_descriptorBufferProperties.samplerDescriptorSize;
-                // fall through
+                [[fallthrough]];
             case VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER:
[...]

PR #139, opened 2026-01-05
common: address clang compiler warnings

crmurillo:
This patch fixes clang compiler warnings as proposed in Issue #134 and adds support in CI to build linux and windows targets using clang.

The clang version used is 21.1.7 and the following warnings were fixed:

```
vulkan-video-samples/vk_video_decoder/libs/NvVideoParser/src/VulkanAV1Decoder.cpp:2374:9: error: unannotated fall-through between switch labels
```
```
vulkan-video-samples/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h:275:13: error: unannotated fall-through between switch labels [-Werror,-Wimplicit-fallthrough]
```
```
vulkan-video-samples/common/libs/VkCodecUtils/DecoderConfig.h:117:18: error: lambda capture 'argv' is not used [-Werror,-Wunused-lambda-capture]
```
```
vulkan-video-samples/common/libs/VkCodecUtils/VulkanVideoProcessor.h:126:26: error: private field 'm_settings' is not used [-Werror,-Wunused-private-field]
```
```
vulkan-video-samples/vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.h:332:14: error: private field 'm_useSeparateLinearImages' is not used [-Werror,-Wunused-private-field]
```
Additional changes suggested in Issue #134 were also considered even though not reproduced:
 - Removal of unused headers
 - Fix preprocessor directive (_WIN32 -> defined(_WIN32))

### DIFF
--- a/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
+++ b/common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
@@ -19,7 +19,6 @@
 
 #include <sstream>
 #include <vulkan_interfaces.h>
-#include "VkCodecUtils/Helpers.h"
 #include "VkCodecUtils/VulkanDeviceContext.h"
 #include "VkCodecUtils/VkBufferResource.h"
 #include "VkCodecUtils/VulkanSamplerYcbcrConversion.h"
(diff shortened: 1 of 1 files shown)


vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.h
```diff
@@ -329,7 +323,7 @@ class VkVideoDecoder : public IVulkanVideoDecoderHandler {
     uint32_t m_useImageViewArray : 1;
     uint32_t m_useSeparateOutputImages : 1;
     uint32_t m_useLinearOutput : 1;
-    uint32_t m_useSeparateLinearImages : 1;
+    [[maybe_unused]] uint32_t m_useSeparateLinearImages : 1;
```
dabrain34: I would remove it as it is unused since 056c211b549412191c182052fb1b2a5114dee5de

common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
```diff
@@ -271,13 +270,13 @@ class VulkanDescriptorSetLayout {
             switch (pDescriptorWrite->descriptorType) {
             case VK_DESCRIPTOR_TYPE_SAMPLER:
                 descriptorSize = m_descriptorBufferProperties.samplerDescriptorSize;
-                // fall through
+                [[fallthrough]];
```
dabrain34: In the following [commit](3a7725ec41ebdeb09f1c606cb75a45d3680bc56a), there was already the use of `[[fallthrough]];[[fallthrough]];` but it has been reverted. It is not clear why, maybe due to MSVC not supporting it.

Would be nice to run CI to check that the change https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/139/files#diff-c790c4ed2a11213f927ae55e2d83935be5173f7efdcf99d698ce5457c956b3ebR2371 is passing as well on Windows.


But an another hand i dont understand why there is no break instead here, as this is suspicious that `descriptorSize` value will be always the latest value `m_descriptorBufferProperties.sampledImageDescriptorSize;` for `VK_DESCRIPTOR_TYPE_SAMPLER` and `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`

@zlatinski any idea on what happens here ?

dabrain34 (changes requested):

common/libs/VkCodecUtils/VulkanDescriptorSetLayout.h
zlatinski: "// fall through" should be supported and undestood by older and newer compilers

[[fallthrough]] is C++17 only and the Vulkan Video samples  supports C++14 and even C++11.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.h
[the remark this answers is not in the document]
crmurillo: The reason why the CI jobs are part of this MR is to ensure all Clang warnings are fixed.
All the changes related to CI are isolated in a separate commit , so they are easy to identify.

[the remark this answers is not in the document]
zlatinski: OK, sounds good. Thank you for clarifying this Stephane!

PR #141, opened 2026-01-07
Add clang compiler support for Linux and Windows builds

crmurillo:
This patch adds support in CI for Linux and Windows builds using clang.

- Add linux-x86_64-clang and linux-x86-clang platforms using clang/clang++.
- Add windows-x64-clang and windows-x86-clang platforms using ClangCL toolset.
- Update CMakeLists.txt to detect clang-cl and use gcc style flags.
- Add #include <intrin.h> for Windows in cpudetect.h. Fixes clang-cl an implicit declaration warning for _BitScanForward and _BitScanForward64.


dabrain34:
maybe it would be nice to have this MR merged with #139

crmurillo:
The changes in this PR are now part of #139.

PR #171, opened 2026-02-17
encode: codec profile and level fixes

srinathkr-nv:
## Description

This series of commits refactors the code related to the determination of the codec profile and level (and tier, as applicable) for all codecs, to avoid redundant calculations / storage and to provide a single point early in the encoder initialization stage where these values are decided.

Another effect of this refactoring is to use the proper codec profile when creating the video profile used for queries of capabilities and recommended settings. Previously, the video profile constructed for these purposes used a default codec profile, which may not be representative of (or may not support) the coding tools requested.

## Type of change

Bug fixes and refactoring

### NVIDIA L2 / NVIDIA 580.94.17 / Ubuntu 24.04.3 LTS

Total Tests:    70
Passed:         50
Crashed:         0
Failed:          0
Not Supported:   2
Skipped:        18 (in skip list)
Success Rate: 100.0%

Prior to this series of commits, running the test framework would show 1 unsupported test case, and 51 passing test cases. The unsupported test case was `encode_h265_main10_profile`, which is actually supported by the NVIDIA driver. The list of passing tests included `encode_av1_high_profile` and `encode_av1_professional_profile` although these profiles are not supported.

With the current changes, `encode_av1_high_profile` and `encode_av1_professional_profile` are correctly marked as unsupported and `encode_h265_main10_profile` executes successfully.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
@@ -285,6 +285,29 @@ VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkD
     return VK_SUCCESS;
 }
 
+void EncoderConfigAV1::InitProfileLevel()
+{
+    // If profile hasn't been specified, determine it based on bit depth and chroma
+    if (profile == STD_VIDEO_AV1_PROFILE_INVALID) {
+        // PROFESSIONAL is required for 12-bit or 422
+        if ((input.bpp > 10) ||
+            (encodeChromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_422_BIT_KHR)) {
+            profile = STD_VIDEO_AV1_PROFILE_PROFESSIONAL;
+        }
+        // HIGH is required for 444 chroma
+        else if (encodeChromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR) {
+            profile = STD_VIDEO_AV1_PROFILE_HIGH;
+        }
+        // MAIN supports 8-bit and 10-bit with 420
+        else {
+            profile = STD_VIDEO_AV1_PROFILE_MAIN;
+        }
+    }
+
+    // Determine level and tier based on encoder configuration
+    DetermineLevelTier();
+}
+
 int8_t EncoderConfigAV1::InitDpbCount()
 {
     dpbCount = STD_VIDEO_AV1_NUM_REF_FRAMES + 1; // BUFFER_POOL_MAX_SIZE = Number of frames in buffer pool = 10
@@ -351,8 +374,7 @@ bool EncoderConfigAV1::DetermineLevelTier()
 
 bool EncoderConfigAV1::InitRateControl()
 {
-    DetermineLevelTier();
-
+    // Level and tier are already initialized by InitProfileLevel()
     // use level max values for now. Limit it to 120Mbits/sec
     uint32_t levelBitrate = std::min(GetLevelBitrate(level, tier), 120000000u);
 
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
@@ -199,8 +199,10 @@ bool EncoderConfigAV1::InitSequenceHeader(StdVideoAV1SequenceHeader *seqHdr,
     return true;
 }
 
-VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkDevCtx)
+VkResult EncoderConfigAV1::InitVideoProfileCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    videoCoreProfile = MakeVideoProfile(static_cast<uint32_t>(profile));
+
     const bool feedback2Requested = enablePictureFeedback || enablePixelCountFeedback ||
                                     enableSkippedPixelCountFeedback || enablePerPartitionFeedback;
     videoEncodeFeedback2Capabilities = VkVideoEncodeFeedback2CapabilitiesKHR
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
@@ -201,6 +201,24 @@ bool EncoderConfigAV1::InitSequenceHeader(StdVideoAV1SequenceHeader *seqHdr,
 
 VkResult EncoderConfigAV1::InitVideoProfileCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    // If profile hasn't been specified, determine it based on bit depth and chroma subsampling
+    if (profile == STD_VIDEO_AV1_PROFILE_INVALID) {
+        // PROFESSIONAL is required for 12-bit or 422
+        if ((encodeBitDepthLuma > 10) ||
+            (encodeChromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_422_BIT_KHR)) {
+            profile = STD_VIDEO_AV1_PROFILE_PROFESSIONAL;
+        }
+        // HIGH is required for 444 chroma
+        else if (encodeChromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR) {
+            profile = STD_VIDEO_AV1_PROFILE_HIGH;
+        }
+        // MAIN supports 8-bit and 10-bit with 420
+        else {
+            profile = STD_VIDEO_AV1_PROFILE_MAIN;
+        }
+    }
+
+    assert(profile != STD_VIDEO_AV1_PROFILE_INVALID);
     videoCoreProfile = MakeVideoProfile(static_cast<uint32_t>(profile));
 
     const bool feedback2Requested = enablePictureFeedback || enablePixelCountFeedback ||
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
@@ -376,17 +376,14 @@ bool EncoderConfigAV1::DetermineLevelTier()
         }
     }
     if (lvl > STD_VIDEO_AV1_LEVEL_7_3) {
-        level = STD_VIDEO_AV1_LEVEL_7_3;
-        tier = 0;
+        return false;
     }
-
     return true;
 }
 
 bool EncoderConfigAV1::InitRateControl()
 {
-    DetermineLevelTier();
-
+    // Level and tier are already initialized by DetermineLevelTier()
     // use level max values for now. Limit it to 120Mbits/sec
     uint32_t levelBitrate = std::min(GetLevelBitrate(level, tier), 120000000u);
 
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -63,6 +63,13 @@ StdVideoH264LevelIdc EncoderConfigH264::DetermineLevel(uint8_t dpbSize,
                                                        uint32_t _vbvBufferSize,
                                                        double frameRate)
 {
+    uint32_t cpbBrNalFactor = 1200;
+
+    // Adjust cpbBrNalFactor depending on whether the High (or greater) profile
+    // is being used
+    if (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH) {
+        cpbBrNalFactor = (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE) ? 4800 : 1500;
+    }
 
     uint32_t frameSizeInMbs = pic_width_in_mbs * pic_height_in_map_units;
     for (uint32_t idx = 0; idx < levelLimitsSize; idx++) {
@@ -71,8 +78,8 @@ StdVideoH264LevelIdc EncoderConfigH264::DetermineLevel(uint8_t dpbSize,
         if ((frameSizeInMbs) > ((uint32_t)levelLimits[idx].maxFS)) continue;
         if ((frameSizeInMbs * numRefFrames * 384) > levelLimits[idx].maxDPB * 1024) continue;
 
-        if ((bitrate != 0) && (bitrate > ((uint32_t)levelLimits[idx].maxBR * 1200))) continue;
-        if ((_vbvBufferSize != 0) && (_vbvBufferSize > ((uint32_t)levelLimits[idx].maxCPB * 1200))) continue;
+        if ((bitrate != 0) && (bitrate > ((uint32_t)levelLimits[idx].maxBR * cpbBrNalFactor))) continue;
+        if ((_vbvBufferSize != 0) && (_vbvBufferSize > ((uint32_t)levelLimits[idx].maxCPB * cpbBrNalFactor))) continue;
 
         return levelLimits[idx].level;
     }
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -505,28 +505,18 @@ int8_t EncoderConfigH264::InitDpbCount()
 
     assert(pic_width_in_mbs > 0);
     assert(pic_height_in_map_units > 0);
-    uint32_t frameSizeInMbs = pic_width_in_mbs * pic_height_in_map_units;
 
     double frameRate = ((frameRateNumerator > 0) && (frameRateDenominator > 0))
                            ? (double)frameRateNumerator / frameRateDenominator
                            : (double)FRAME_RATE_NUM_DEFAULT / (double)FRAME_RATE_DEN_DEFAULT;
 
-    // WAR for super HD resolution (bypass H264 level check for frame mode and use level 5.2)
-    if ((frameSizeInMbs > ((uint32_t)levelLimits[levelLimitsSize - 1].maxFS) ||
-        ((frameSizeInMbs * frameRate) > ((uint32_t)levelLimits[levelLimitsSize - 1].maxMBPS)))) {
-        levelIdc = STD_VIDEO_H264_LEVEL_IDC_5_2;
-    } else {
-        // find lowest possible level
-        levelIdc = DetermineLevel(dpbCount, levelBitRate, vbvBufferSize, frameRate);
-    }
+    // find lowest possible level
+    levelIdc = DetermineLevel(dpbCount, levelBitRate, vbvBufferSize, frameRate);
 
     uint8_t levelDpbSize = (uint8_t)(((1024 * levelLimits[levelIdc].maxDPB)) /
                             ((pic_width_in_mbs * pic_height_in_map_units) * 384));
 
-    // XXX: If the level is 5.2, it is highly likely that we forced it to that
-    // value as a WAR for super HD. In that case, force the DPB size to
-    // DEFAULT_MAX_NUM_REF_FRAMES. Otherwise, clamp the computed DPB size to DEFAULT_MAX_NUM_REF_FRAMES.
-    levelDpbSize = (levelIdc == STD_VIDEO_H264_LEVEL_IDC_5_2) ? (uint8_t)DEFAULT_MAX_NUM_REF_FRAMES : std::min(uint8_t(DEFAULT_MAX_NUM_REF_FRAMES), levelDpbSize);
+    levelDpbSize = std::min(uint8_t(DEFAULT_MAX_NUM_REF_FRAMES), levelDpbSize);
 
     uint8_t dpbSize = (uint8_t)((dpbCount < 1) ? levelDpbSize : (uint8_t)(std::min((uint8_t)dpbCount, levelDpbSize))) + 1;
 
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -310,11 +310,6 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
         sps->pic_order_cnt_type = STD_VIDEO_H264_POC_TYPE_2;
     }
 
-    // FIXME: Check if the HW supports transform_8x8_mode_is_supported
-    // based on capabilities or profiles supported
-    const bool transform_8x8_mode_is_supported = true;
-    const bool bIsFastestPreset = false;
-
     if (adaptiveTransformMode == ADAPTIVE_TRANSFORM_ENABLE) {
         pps->flags.transform_8x8_mode_flag = true;
         if ((profileIdc == STD_VIDEO_H264_PROFILE_IDC_BASELINE) ||
@@ -326,12 +321,9 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
         pps->flags.transform_8x8_mode_flag = false;
     } else {
         // Autoselect
-        if (!bIsFastestPreset || transform_8x8_mode_is_supported) {
-            if ((profileIdc == STD_VIDEO_H264_PROFILE_IDC_INVALID) ||
-                (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH)) {
-                // Unconditionally enable 8x8 transform
-                pps->flags.transform_8x8_mode_flag = true;
-            }
+        if (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH) {
+            // Unconditionally enable 8x8 transform
+            pps->flags.transform_8x8_mode_flag = true;
         }
     }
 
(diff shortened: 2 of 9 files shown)


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
```diff
@@ -285,6 +285,29 @@ VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkD
     return VK_SUCCESS;
 }
 
+void EncoderConfigAV1::InitProfileLevel()
```
dabrain34: This method `InitProfileLevel` is called too early preventing `DetermineLevelTier` to work properly because `numRefFrames` member variable line 79 will be always 0 at this time. 

I would recommend to calculate all of that after having called `InitDeviceCapabilities` to get a valid `numRefFrames`.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -71,8 +78,8 @@ StdVideoH264LevelIdc EncoderConfigH264::DetermineLevel(uint8_t dpbSize,
         if ((frameSizeInMbs) > ((uint32_t)levelLimits[idx].maxFS)) continue;
         if ((frameSizeInMbs * numRefFrames * 384) > levelLimits[idx].maxDPB * 1024) continue;
```
dabrain34, line 80: This method `DetermineLevel` is called too early preventing  to work as expected because `numRefFrames` member variable  will be always 0 at this time. 

I would recommend to calculate all of that after having called `InitDeviceCapabilities` to get a valid `numRefFrames`.

This can be performed by calling `InitProfileLevel` in `InitEncoder`

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -491,42 +468,67 @@ VkResult EncoderConfigH264::InitDeviceCapabilities(const VulkanDeviceContext* vk
...
-    // WAR for super HD resolution (bypass H264 level check for frame mode and use level 5.2)
-    if ((frameSizeInMbs > ((uint32_t)levelLimits[levelLimitsSize - 1].maxFS) ||
-        ((frameSizeInMbs * frameRate) > ((uint32_t)levelLimits[levelLimitsSize - 1].maxMBPS)))) {
-        levelIdc = STD_VIDEO_H264_LEVEL_IDC_5_2;
-    } else {
-        // find lowest possible level
-        levelIdc = DetermineLevel(dpbCount, levelBitRate, vbvBufferSize, frameRate);
-    }
+    // find lowest possible level
+    levelIdc = DetermineLevel(dpbCount, levelBitRate, vbvBufferSize, frameRate);
```
dabrain34, line 561: DetermineLevel might return STD_VIDEO_H264_LEVEL_IDC_INVALID which can lead to incorrect access to `levelLimits` array. I would recommend to display an error and use a valid value or return a proper error in `VkVideoEncoder::InitEncoder` if `InitProfileLevel` fails to initialize

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -483,53 +483,67 @@ void EncoderConfigH265::InitializeSpsRefPicSet(SpsH265 *pSps)
...
 {
-    StdVideoH265ProfileTierLevel profileTierLevel{};
-    profileTierLevel.general_profile_idc = STD_VIDEO_H265_PROFILE_IDC_INVALID;
-    profileTierLevel.general_level_idc = STD_VIDEO_H265_LEVEL_IDC_INVALID;
+    levelIdc = STD_VIDEO_H265_LEVEL_IDC_INVALID;
+    general_tier_flag = 0;
     uint32_t levelIdx = 0;
     for (; levelIdx < levelLimitsTblSize; levelIdx++) {
 
         if (IsSuitableLevel(levelIdx, 0)) {
```
dabrain34, line 538: IsSuitableLevel might returns earlier due to the fact that GetCpbVclFactor is wrongly computed due to initalize value such as encodeBitDepthChroma or encodeBitDepthLuma which will e initialized in `InitVideoProfile` so it should be called after  such as in https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/changes#r3110857666

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.h
```diff
@@ -83,8 +83,8 @@ struct EncoderConfigH265 : public EncoderConfig {
     size_t                 levelLimitsTblSize;
 
     EncoderConfigH265()
-      : profile(STD_VIDEO_H265_PROFILE_IDC_MAIN)
-      , levelIdc(STD_VIDEO_H265_LEVEL_IDC_5_2)
+      : profile(STD_VIDEO_H265_PROFILE_IDC_INVALID)
```
dabrain34, line 88: the last branch of `levelLimitsTbl` is unreachable as there is no stdLevelIdc for this branch

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -326,12 +328,9 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
         // Autoselect
-        if (!bIsFastestPreset || transform_8x8_mode_is_supported) {
-            if ((profileIdc == STD_VIDEO_H264_PROFILE_IDC_INVALID) ||
-                (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH)) {
-                // Unconditionally enable 8x8 transform
-                pps->flags.transform_8x8_mode_flag = true;
-            }
+        if (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH) {
+            // Unconditionally enable 8x8 transform
+            pps->flags.transform_8x8_mode_flag = true;
```
dabrain34: This value has been already computed in use8x8Transform maybe we could keep this value and reuse it or have a method to calculate properly this value. I would be in favor to have a member value

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -63,6 +63,13 @@ StdVideoH264LevelIdc EncoderConfigH264::DetermineLevel(uint8_t dpbSize,
                                                        uint32_t _vbvBufferSize,
                                                        double frameRate)
 {
+    uint32_t cpbBrNalFactor = 1200;
+
+    // Adjust cpbBrNalFactor depending on whether the High (or greater) profile
+    // is being used
+    if (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH) {
+        cpbBrNalFactor = (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE) ? 4800 : 1500;
```
dabrain34, line 72: would be nice to explain these values with the spec reference

dabrain34 (changes requested):

dabrain34:
@srinathkr-nv ping :)

srinathkr-nv:
Sorry for the delay; I have previously noticed your comments but I haven't been able to get around to addressing them. I'll try doing that next week.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -491,42 +468,67 @@ VkResult EncoderConfigH264::InitDeviceCapabilities(const VulkanDeviceContext* vk
     return VK_SUCCESS;
 }
 
-int8_t EncoderConfigH264::InitDpbCount()
+void EncoderConfigH264::InitProfileLevel()
 {
-    dpbCount = 0; // TODO: What is the need for this?
+    // FIXME: Check if the HW supports transform_8x8_mode_is_supported
```
dabrain34: could we address this FIXME because in the case of drivers which does not support HIGH profile, the config defaults on HIGH profile and it is failing

dabrain34: could we set by default this use8x8Transform to false as the deviceCapabilities are not queried yet and the hardware might not support it ?

dabrain34: here is a fix for the issue: https://github.com/dabrain34/Vulkan-Video-Samples/commit/15085361e37290156617067a475890e56cd3e1da you can cherry pick it if you approve the change :)

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
srinathkr-nv: I have now moved both the profile and level determination code closer to the device capabilities query.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: I have attempted to address this in my latest set of commits; can you review them?

I'm also working on testing it more and addressing some of the other edge cases you mentioned above.

srinathkr-nv:
I have added new commits to this PR to address some of the main issues with the profile and level determination logic. This essentially reverses the direction I had taken in the earlier commits in this series, so I will be reworking this set of changes again to have a clean history.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: I have addressed this in the previous batch of commits (which moved both the profile and level determination code closer to the device capabilities query).

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: Done in [d8fcc9c](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/d8fcc9c3fdd1ae83b92c5611fcd56392584f4041)

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: I have addressed this in [0fe29b3](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/0fe29b346c1b1ff7bb854a58fc64703519a585e0)

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
srinathkr-nv: This should be addressed now by my earlier batch of commits (which moved both the profile and level determination code closer to the device capabilities query).

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: Good point; I have attempted to address this in [e9599e9](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/e9599e95ef9a2d4a7b667074f4bc680772e1e477)

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.h
srinathkr-nv: I'm not sure what this is referring to; can you clarify?

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -310,29 +317,12 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
-        if ((profileIdc == STD_VIDEO_H264_PROFILE_IDC_BASELINE) ||
-            (profileIdc == STD_VIDEO_H264_PROFILE_IDC_MAIN)) {
-            fprintf(stderr, "The profile selected does not support transform_8x8_mode_flag, disabling it.\n");
-            pps->flags.transform_8x8_mode_flag = false;
-        }
-    } else if (adaptiveTransformMode == ADAPTIVE_TRANSFORM_DISABLE) {
-        pps->flags.transform_8x8_mode_flag = false;
+        pps->flags.transform_8x8_mode_flag = (profileIdc >= STD_VIDEO_H264_PROFILE_IDC_HIGH);
+        if (!pps->flags.transform_8x8_mode_flag)
+            fprintf(stderr, "The transform_8x8_mode_flag has been disabled because the selected profile selected does not support it.\n");
```
dabrain34: you can remove the duplicate `selected`

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -626,7 +647,9 @@ bool EncoderConfigH265::InitParamameters(VpsH265 *vpsInfo, SpsH265 *spsInfo,
         spsInfo->decPicBufMgr.max_num_reorder_pics[i] = gopStructure.GetConsecutiveBFrameCount() ? 1 : 0;
     }
 
-    spsInfo->profileTierLevel = GetLevelTier();
+    // Use pre-initialized profile, level, and tier from InitCodecProfile() / DetermineLevelTier()
```
dabrain34: should be `InitVideoProfileCapabilities` instead of `InitCodecProfile`

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -483,53 +510,47 @@ void EncoderConfigH265::InitializeSpsRefPicSet(SpsH265 *pSps)
...
     if (levelIdx >= levelLimitsTblSize) {
-        assert(!"No suitable level selected");
+        return false;
     }
-
-    return profileTierLevel;
+    return (levelIdc <= h265EncodeCapabilities.maxLevelIdc);
 }
 
+
```
dabrain34: extra line

dabrain34:
Thanks for this set of patches, I confirm that capabilities is retrieved before determining the right profile and level. 

Could you please squash the commit history in order to avoid the intermediate non working state ?

srinathkr-nv:
> Thanks for this set of patches, I confirm that capabilities is retrieved before determining the right profile and level.
> 
> Could you please squash the commit history in order to avoid the intermediate non working state ?

Great, thank you for trying it out and confirming that it works for you. I was waiting for this confirmation before I reorganized the commits; I will work on this part next.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: Fixed in the latest series of commits.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
srinathkr-nv: Fixed in the latest series of commits.

dabrain34:
That's great news I will have a deeper look a bit later.
I feel that we could still squash some commits such as the h264 level ones or the h264config ones as 10 commits looks quite a lot in that PR.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -498,53 +528,46 @@ void EncoderConfigH265::InitializeSpsRefPicSet(SpsH265 *pSps)
...
         }
     }
 
     if (levelIdx >= levelLimitsTblSize) {
-        assert(!"No suitable level selected");
+        return false;
     }
-
-    return profileTierLevel;
+    return (levelIdc <= h265EncodeCapabilities.maxLevelIdc);
```
crmurillo, line 556: I've been running some tests for 10-bit support and needed the changes in this PR to run the test encode_h265_main10_profile. While testing this PR on radv (RX7600/VCN 4.0, Mesa 26.2), DetermineLevelTier() fails because radv reports maxLevelIdc = 0 for H265 encode. In Mesa, there are static tables that say level 6.1, but it seems they are overwritten by the kernel's AMDGPU_INFO_VIDEO_CAPS_ENCODE max_level, which is hardcoded to 0.
@lolzballs: do you know where maxLevelIdc should come from? a constant pre-determined by the hardware or queried at runtime (kernel via ioctl)?

lolzballs: Thanks, https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43584 should fix it.

dabrain34:
@srinathkr-nv could you please review the history of commits and shorten a bit ? See my comment https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171#issuecomment-5036337688

dabrain34 (changes requested):
please squash some commits

srinathkr-nv:
I have now folded four of the original commits which touched the the H.264 level determination code into a single commit.

The commits [a409e1f](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/a409e1f5b859060cef9d40ac7221221c5e9af28b), [5886e91](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/5886e91d64b36082b791e023f10e13c2db0a1f17), [c8863a2](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/c8863a244e673b7331dfad5060fda7692e408a79) and [1827c29](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171/commits/1827c29d6d229ae637e16362750146ea64464d6b) fundamentally re-architect the code involving profile and level determination and should be kept as they are.

That leaves the remaining two commits in this PR, which are intentionally kept separate because it is awkward to squash them into any of the other commits (and they have a very specific focus area).

dabrain34:
We dont use `Signed-off` tag in this repository in the commit message, any reason you'd like to have it now ?
Regarding the commit message, it can be very verbose and pretty difficult to read.

Regarding the history, the 4 patches does not exist anymore, so I dont know what are the patches you'd like to keep.

I would squash `02cc3936 EncoderConfigH264: add a profile-probing loop in InitVideoProfileCapabilities` and `31652193 EncoderConfigH264: simplify transform_8x8_mode_flag handling` into 
`4b68dcf8 encode: add/move codec profile auto-selection into InitVideoProfileCapabilities` as this is part of improving the auto selection of profile.



`EncoderConfigH265: add MaxLumaSr check to IsSuitableLevel` and `4824c3e7 VkEncoderConfigH264: improve level limits table and related code` could also be squashed in  `088ba79a encode: separate level/tier determination into a distinct initialization step`

And then clarify a bit the commit message that we are improving the level selection and then the profile selection.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -391,28 +359,101 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
+    if (autoSelected) {
+        // Select the highest profile required for the requested features.
+        // 8x8 transform support is hardware-dependent; it is verified after
+        // the capability query and the mode is adjusted accordingly.
+        bool use8x8Transform = (adaptiveTransformMode != ADAPTIVE_TRANSFORM_DISABLE);
+
+        profileIdc = STD_VIDEO_H264_PROFILE_IDC_BASELINE;
+
+        // Upgrade to MAIN profile if using B-frames or CABAC entropy coding
+        if ((gopStructure.GetConsecutiveBFrameCount() > 0) || (entropyCodingMode == ENTROPY_CODING_MODE_CABAC))
```
dabrain34, line 386: by default GetConsecutiveBFrameCount will be initialized to CONSECUTIVE_B_FRAME_COUNT_MAX_VALUE which is 255. So I would first validate that you want BFrameCount with the command line in order to select the right profile according to this. Otherwise it will never be STD_VIDEO_H264_PROFILE_IDC_BASELINE

you could for example add the line 

```
const uint8_t bFrameCount = gopStructure.GetConsecutiveBFrameCount();
const bool wantsBFrames = (bFrameCount != CONSECUTIVE_B_FRAME_COUNT_MAX_VALUE) && (bFrameCount > 0);
```

And use `wantsBFrames` to validate that b-frame has been selected

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -391,28 +359,101 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
+            profileIdc = STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE;
+        }
+    }
+
+    // Determine the minimum profile required by the content format.
+    StdVideoH264ProfileIdc minProfile = STD_VIDEO_H264_PROFILE_IDC_BASELINE;
+    if ((tuningMode == VK_VIDEO_ENCODE_TUNING_MODE_LOSSLESS_KHR) ||
+        (input.chromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR)) {
+        minProfile = STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE;
+    } else if (gopStructure.GetConsecutiveBFrameCount() > 0) {
```
dabrain34, line 405: you should validate that b-frame has been validated by command line argument and not CONSECUTIVE_B_FRAME_COUNT_MAX_VALUE

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -391,28 +359,101 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
+            (input.chromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR)) {
+            profileIdc = STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE;
+        }
+    }
+
+    // Determine the minimum profile required by the content format.
+    StdVideoH264ProfileIdc minProfile = STD_VIDEO_H264_PROFILE_IDC_BASELINE;
+    if ((tuningMode == VK_VIDEO_ENCODE_TUNING_MODE_LOSSLESS_KHR) ||
+        (input.chromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR)) {
+        minProfile = STD_VIDEO_H264_PROFILE_IDC_HIGH_444_PREDICTIVE;
```
dabrain34, line 404: Here STD_VIDEO_H264_PROFILE_IDC_HIGH should be selected if 
f ((encodeBitDepthLuma > 8) ||
   (encodeChromaSubsampling != VK_VIDEO_CHROMA_SUBSAMPLING_420_BIT_KHR))

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
```diff
@@ -391,28 +359,101 @@ bool EncoderConfigH264::InitSpsPpsParameters(StdVideoH264SequenceParameterSet *s
...
+        if ((gopStructure.GetConsecutiveBFrameCount() > 0) || (entropyCodingMode == ENTROPY_CODING_MODE_CABAC))
+            profileIdc = STD_VIDEO_H264_PROFILE_IDC_MAIN;
+
+        if (use8x8Transform) {
+            profileIdc = STD_VIDEO_H264_PROFILE_IDC_HIGH;
+        }
+
+        // Upgrade to HIGH_444_PREDICTIVE for lossless encoding or 4:4:4 chroma
+        if ((tuningMode == VK_VIDEO_ENCODE_TUNING_MODE_LOSSLESS_KHR) ||
+            (input.chromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_444_BIT_KHR)) {
```
dabrain34, line 395: to be consistent with other configs `encodeChromaSubsampling` should be used here instead of `input.chromaSubsampling`.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -112,8 +112,27 @@ int EncoderConfigH265::DoParseArguments(int argc, const char* argv[])
     return 0;
 }
 
-VkResult EncoderConfigH265::InitDeviceCapabilities(const VulkanDeviceContext* vkDevCtx)
+VkResult EncoderConfigH265::InitVideoProfileCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    // If profile hasn't been specified, determine it based on bit depth and chroma subsampling
+    if (profile == STD_VIDEO_H265_PROFILE_IDC_INVALID) {
+        if (encodeChromaSubsampling == VK_VIDEO_CHROMA_SUBSAMPLING_420_BIT_KHR) {
+            if (input.bpp == 8) {
```
dabrain34, line 120: it should use `encodeBitDepthLuma`

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
```diff
@@ -188,8 +188,28 @@ bool EncoderConfigAV1::InitSequenceHeader(StdVideoAV1SequenceHeader *seqHdr,
     return true;
 }
 
-VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkDevCtx)
+VkResult EncoderConfigAV1::InitVideoProfileCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    // If profile hasn't been specified, determine it based on bit depth and chroma subsampling
+    if (profile == STD_VIDEO_AV1_PROFILE_INVALID) {
+        // PROFESSIONAL is required for 12-bit or 422
+        if ((input.bpp > 10) ||
```
dabrain34, line 196: it should use `encodeBitDepthLuma`

dabrain34 (changes requested):

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: I have fixed this in the latest series of commits.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
srinathkr-nv: I have fixed this in the latest series of commits.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
srinathkr-nv: I have fixed this in the latest series of commits.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: I will see if there's a better way to do this; if not, I'll apply your suggestion to the code.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
srinathkr-nv: The `HIGH` profile for H.264 does not support bit-depth > 8 and a chroma subsampling other than 4:2:0 and 4:0:0. The correct profile for those cases would be the `HIGH_10` and `HIGH_422` profiles, but correctly using those profiles will require additional changes beyond just the profile- and level-related fixes which I'm making here.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
dabrain34: any progress on this ?

dabrain34: what do you think about merging this MR and fix this a bit later ?

srinathkr-nv: I had implemented this approach yesterday locally (folded into the appropriate prior commits) but I didn't push the updates as I hadn't re-rerun the tests. I have created https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/238 now with the fix for the above issue.

PR #186, opened 2026-03-20
demuxer: Fix demuxer assert logic and refactor FindFFmpeg

crmurillo:
## Description

This PR addresses two issues:

1. Logic inside the demuxer to detect errors in case FFmpeg is not found.

When FFmpeg is not found at build time and no codec is explicitly passed at runtime, codecType defaults to VK_VIDEO_CODEC_OPERATION_NONE_KHR. The assert() calls validating this are stripped in release builds, so the demuxer silently proceeds with invalid parameters and fails with a misleading "Video decode queue family not supported" error. See linked issue for more details.

2. The cmake logic to detect FFmpeg and the relevant required libraries.

The FindFFmpeg.cmake module had two separate search passes for FFmpeg libraries: one via a custom FFMPEG_FIND macro and another via pkg_check_modules + find_library. It used FFMPEG as the pkg-config prefix, which caused pkg-config to overwrite the module's own output variables 

## Type of change

bug fix

## Issue (optional)

Fixes #185.

### NVIDIA GeForce RTX 4060 Ti (10de:2803) / NVIDIA 595.44.00 / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         73 (1 warning(s))
Crashed:         0
Failed:          0
Not Supported:  11
Success Rate: 100.0%

### AMD Radeon RX 7600 (RADV NAVI33) (1002:7480) / radv Mesa 26.3.0-devel (git-fe1491b758) / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  12
Success Rate: 100.0%

### Intel(R) UHD Graphics 770 (ADL-S GT1) (8086:4680) / Intel open-source Mesa driver Mesa 26.3.0-devel (git-fe1491b758) / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         57
Crashed:         0
Failed:          0
Not Supported:  25
Success Rate: 100.0%

### AMD Radeon RX 6600 / AMD proprietary driver 26.3.1 (AMD proprietary shader compiler) / Windows 10

Total Tests:    73
Passed:         37
Crashed:         0
Failed:          0
Not Supported:  28
Skipped:         8 (in skip list)
Success Rate: 100.0%


vk_video_decoder/demos/vk-video-dec/Main.cpp
```diff
@@ -45,6 +45,9 @@ int main(int argc, const char **argv)
     if (result != VK_SUCCESS) {
         fprintf(stderr, "Error: Failed to initialize VideoStreamDemuxer for file: %s\n",
                 decoderConfig.videoFileName.c_str());
+#ifndef FFMPEG_DEMUXER_SUPPORT
+        fprintf(stderr, "Note: This build does not include FFmpeg demuxer support.\n");
```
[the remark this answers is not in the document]
crmurillo: Log moved to `VideoStreamDemuxer::Create` and redirected to stdout.

vk_video_decoder/test/vulkan-video-simple-dec/Main.cpp
```diff
@@ -141,6 +141,9 @@ int main(int argc, const char** argv)
     if (result != VK_SUCCESS) {
         fprintf(stderr, "Error: Failed to initialize VideoStreamDemuxer for file: %s\n",
                 decoderConfig.videoFileName.c_str());
+#ifndef FFMPEG_DEMUXER_SUPPORT
+        fprintf(stderr, "Note: This build does not include FFmpeg demuxer support.\n");
```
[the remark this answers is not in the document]
crmurillo: Log moved to VideoStreamDemuxer::Create and redirected to stdout.

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,125 @@
...
+#   FFMPEG_INCLUDE_DIR             - FFmpeg include directory
+#   FFMPEG_INCLUDE_DIRS            - Same as above (alias)
+#   FFMPEG_LIBRARIES               - All FFmpeg libraries
+#   FFMPEG_LIBAVCODEC_LIBRARIES    - avcodec library
+#   FFMPEG_LIBAVFORMAT_LIBRARIES   - avformat library
+#   FFMPEG_LIBAVUTIL_LIBRARIES     - avutil library
+#   FFMPEG_LIBSWSCALE_LIBRARIES    - swscale library (optional)
+#
+# Accepts:
+#   FFMPEG_ROOT or ENV{FFMPEG_DIR} or ENV{FFMPEG_ROOT} - Custom search path
```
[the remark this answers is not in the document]
crmurillo: The logic to set the include directory based on the platform is moved to `FindFFmpeg.cmake`, where now both the library and the include paths are set for all platforms.

The `VULKAN_FFMPEG_LIB_PATH` remains in the `CMakeLists.txt` because it is required to install the libraries in case FFmpeg is found.

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,158 @@
...
+    PATHS ${FFMPEG_INCLUDE_SEARCH_PATHS}
     PATH_SUFFIXES ffmpeg
 )
 
-# Find libraries
-find_library(AVCODEC_LIBRARY
+# Common library search paths
+set(FFMPEG_LIB_SEARCH_PATHS
+    $ENV{FFMPEG_DIR}/lib
+    $ENV{FFMPEG_ROOT}/lib
```
dabrain34: why do we need to search in $ENV{FFMPEG_ROOT} ?

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,158 @@
...
+    elseif((CMAKE_GENERATOR_PLATFORM MATCHES "^(arm|ARM)"))
+        set(FFMPEG_WIN32_PREBUILT_DIR "${FFMPEG_WIN32_PREBUILT_ROOT}/winarm")
+    else()
+        set(FFMPEG_WIN32_PREBUILT_DIR "${FFMPEG_WIN32_PREBUILT_ROOT}/win64")
+    endif()
+endif()
+
+# Common include search paths
+set(FFMPEG_INCLUDE_SEARCH_PATHS
+    $ENV{FFMPEG_DIR}/include
```
dabrain34: why do we need to search in $ENV{FFMPEG_ROOT} and $ENV{FFMPEG_DIR}?

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,158 @@
...
+#   FFMPEG_LIBAVFORMAT_LIBRARIES   - avformat library
+#   FFMPEG_LIBAVUTIL_LIBRARIES     - avutil library
+#   FFMPEG_LIBSWSCALE_LIBRARIES    - swscale library (optional)
+#
+# Accepts:
+#   FFMPEG_ROOT or ENV{FFMPEG_DIR} or ENV{FFMPEG_ROOT} - Custom search path
+#
 
-    SET(FFMPEG_LIBRARY_DIRS ${FFMPEG_LIBAVFORMAT_LIBRARY_DIRS})
+set(FFMPEG_ROOT "$ENV{FFMPEG_DIR}" CACHE PATH "Location of FFmpeg")
```
dabrain34: I would move that before the search paths definition to readability and I would document this env var somewhere (thinking about removing it). Is it something you are using @zlatinski  ?

CMakeLists.txt
```diff
@@ -74,18 +74,12 @@ else()
         if ((CMAKE_GENERATOR_PLATFORM MATCHES "^aarch64") OR (CMAKE_GENERATOR_PLATFORM MATCHES "^arm64") OR (CMAKE_GENERATOR_PLATFORM MATCHES "^ARM64"))
```
dabrain34:
```suggestion
  if((CMAKE_GENERATOR_PLATFORM MATCHES "^(aarch64|arm64|ARM64)"))
```

dabrain34: I assume part of this block is necessary but should be moved lower to tell where to install the ffmpeg libraries on Windows.

I think this can be simplified again.

cmake/FindFFmpeg.cmake
crmurillo: IMHO we do not need to search those paths, I just kept all the paths that were in the previous implementation. These variables are not documented anywhere else, so I would be more inclined to remove them.

cmake/FindFFmpeg.cmake
crmurillo: Same as above. 
IMHO we do not need to search those paths, I just kept all the paths that were in the previous implementation. These variables are not documented anywhere else, so I would be more inclined to remove them.

cmake/FindFFmpeg.cmake
crmurillo: I would also prefer to remove it if it is not needed.

CMakeLists.txt
crmurillo: I refactored this logic. It is removed from here as it was duplicating logic in FindFFmpeg.cmake.

cmake/FindFFmpeg.cmake
crmurillo: There are currently 3 environment variables (to keep old behavior):

- FFMPEG_ROOT
- FFMPEG_DIR
- VULKAN_FFMPEG_LIB_DIR_PATH

I added a rough documentation at the top of the FindFFmpeg, but I cannot tell a big difference between them or why we need 3.

cmake/FindFFmpeg.cmake
dabrain34: yes the same, if there are not used and not documented, please remove those

cmake/FindFFmpeg.cmake
dabrain34: I would be in favor to clean this usage of variable

crmurillo: I cleaned up the variables, left only one: FFMPEG_ROOT and documented it.

cmake/FindFFmpeg.cmake
crmurillo: The variables were removed from here, only FFMPEG_ROOT is kept to allow the user to install ffmpeg in a custom path.

cmake/FindFFmpeg.cmake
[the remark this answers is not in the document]
crmurillo: `VULKAN_FFMPEG_LIB_PATH` is removed.

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,156 @@
...
+# Find the native FFmpeg includes and libraries.
+#
+# This module defines:
+#   FFMPEG_FOUND                   - True if FFmpeg was found
+#   FFMPEG_INCLUDE_DIRS            - FFmpeg include directories
+#   FFMPEG_LIBRARIES               - All FFmpeg libraries
+#   FFMPEG_LIBAVCODEC_LIBRARIES    - avcodec library
+#   FFMPEG_LIBAVFORMAT_LIBRARIES   - avformat library
+#   FFMPEG_LIBAVUTIL_LIBRARIES     - avutil library
+#   FFMPEG_LIBSWSCALE_LIBRARIES    - swscale library (optional)
```
dabrain34: As mentioned in https://github.com/KhronosGroup/Vulkan-Video-Samples/issues/185, it is optional I would remove it. I gave it a try on Linux and it works fine

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,156 @@
...
 
-# Mark as advanced
 mark_as_advanced(
-    FFMPEG_INCLUDE_DIR
-    AVCODEC_LIBRARY
-    AVFORMAT_LIBRARY
-    AVUTIL_LIBRARY
-    SWSCALE_LIBRARY
+    ${FFMPEG_REQUIRED_VARS}
+    FFMPEG_LIBSWSCALE_LIBRARIES
```
dabrain34: I would remove it

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,156 @@
...
+        ${FFMPEG_LIBAVUTIL_INCLUDE_DIR}
+    )
+    list(REMOVE_DUPLICATES FFMPEG_INCLUDE_DIRS)
+    set(FFMPEG_LIBRARIES
+        ${FFMPEG_LIBAVCODEC_LIBRARIES}
+        ${FFMPEG_LIBAVFORMAT_LIBRARIES}
+        ${FFMPEG_LIBAVUTIL_LIBRARIES}
+    )
+    if(FFMPEG_LIBSWSCALE_LIBRARIES)
+        list(APPEND FFMPEG_LIBRARIES ${FFMPEG_LIBSWSCALE_LIBRARIES})
```
dabrain34: I would remove it

vk_video_decoder/demos/vk-video-dec/CMakeLists.txt
```diff
@@ -102,11 +102,7 @@ else()
 endif()
 
 if(FFMPEG_AVAILABLE)
-    if(WIN32)
-        list(APPEND libraries PUBLIC ${AVCODEC_LIB} ${AVFORMAT_LIB} ${AVUTIL_LIB})
-    else()
-        list(APPEND libraries PRIVATE ${FFMPEG_LIBAVCODEC_LIBRARIES} ${FFMPEG_LIBAVUTIL_LIBRARIES} ${FFMPEG_LIBAVFORMAT_LIBRARIES})
-    endif()
+    list(APPEND libraries PRIVATE ${FFMPEG_LIBAVCODEC_LIBRARIES} ${FFMPEG_LIBAVFORMAT_LIBRARIES} ${FFMPEG_LIBAVUTIL_LIBRARIES})
```
dabrain34: why not use FFMPEG_LIBRARIES ?

dabrain34 (changes requested):

vk_video_decoder/demos/vk-video-dec/CMakeLists.txt
crmurillo: Fixed here along with other CMakeLists.txt files where FFMPEG_LIBRARIES could be used.

CMakeLists.txt
```diff
@@ -59,85 +59,30 @@ if(BUILD_DECODER OR USE_ENCODER_SHADERC)
...
-                PATTERN "*.a" EXCLUDE)
-
-           install(DIRECTORY "${VULKAN_FFMPEG_LIB_PATH}/"
-                DESTINATION "${CMAKE_INSTALL_PREFIX}/lib")
-       else()
-            message(WARNING "Can't find libavcodec, libavformat, or libavutil on Windows!")
-       endif()
-    else()
-        message(WARNING "Can't find libavcodec, libavformat, or libavutil on Linux!")
+        get_filename_component(FFMPEG_BIN_DIR "${FFMPEG_LIB_DIR}/../bin" ABSOLUTE)
```
dabrain34, line 76: depending on FFMPEG_BIN_DIR contents, you might copy files which are not necessary. You should try to copy only the necessary files coming from ffmpeg package

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,142 @@
...
+# Accepts:
+#   FFMPEG_ROOT or ENV{FFMPEG_ROOT} - Custom search path
+#
 
-# Find FFmpeg components
+# Try pkg-config first
 find_package(PkgConfig QUIET)
 if(PKG_CONFIG_FOUND)
-    pkg_check_modules(FFMPEG QUIET
+    pkg_check_modules(PC_FFMPEG QUIET
```
dabrain34: the package config result should be high priority result setting all the necessary variable. I dont know if all the search and find are necessary afterwhile

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,142 @@
...
+    /usr/lib
+    /usr/lib/aarch64-linux-gnu
+    /usr/lib/x86_64-linux-gnu
+    /usr/lib64
+    /usr/local/lib
+    /usr/local/lib64
+)
+
+# Find required libraries
+find_library(FFMPEG_LIBAVCODEC_LIBRARIES
```
dabrain34: this is kind of confusing to have LIBAR**IES** when we are searching for only one library. can we use `FFMPEG_LIBAVCODEC_LIBRARY` ?

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,142 @@
...
-    NAMES libavcodec/avcodec.h libavformat/avformat.h libavutil/avutil.h libswscale/swscale.h
-    PATHS
-        ${FFMPEG_INCLUDE_DIRS}
-        /usr/include
-        /usr/local/include
-        $ENV{FFMPEG_ROOT}/include
+# Windows FFmpeg paths
+if(WIN32)
+    set(FFMPEG_WIN32_PREBUILT_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/vk_video_decoder/bin/libs/ffmpeg")
+    if((CMAKE_GENERATOR_PLATFORM MATCHES "^(aarch64|arm64|ARM64)"))
```
dabrain34: it seems that CMAKE_GENERATOR_PLATFORM is only used by visual studio generator so it might fail with ninja generation. Better to use CMAKE_SYSTEM_PROCESSOR, see 

https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html

https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html

There is unnecessary double parenthesis

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,142 @@
...
+endif()
+
+set(FFMPEG_ROOT "$ENV{FFMPEG_ROOT}" CACHE PATH "Location of FFmpeg")
+
+# Common include search paths
+set(FFMPEG_INCLUDE_SEARCH_PATHS
+    ${FFMPEG_ROOT}/include
+    ${FFMPEG_WIN32_PREBUILT_DIR}/include
+    ${PC_FFMPEG_INCLUDE_DIRS}
+    /Library/Frameworks
```
dabrain34: it seems that this path wont exist as it is and no binary will be found from here. Can you search in homebrew instead such as /opt/homebrew/

CMakeLists.txt
```diff
@@ -59,85 +59,30 @@ if(BUILD_DECODER OR USE_ENCODER_SHADERC)
...
-# FFMPEG_LIBAVCODEC
-# FFMPEG_LIBAVFORMAT
-# FFMPEG_LIBAVUTIL
 if(FFMPEG_FOUND)
-    message("Found FFMPEG/LibAV libraries")
-    include_directories(${FFMPEG_INCLUDE_DIR})
+    message(STATUS "  libavcodec:  ${FFMPEG_LIBAVCODEC_LIBRARIES}")
+    message(STATUS "  libavformat: ${FFMPEG_LIBAVFORMAT_LIBRARIES}")
+    message(STATUS "  libavutil:   ${FFMPEG_LIBAVUTIL_LIBRARIES}")
+    include_directories(${FFMPEG_INCLUDE_DIRS})
```
dabrain34: this is project wide maybe better to have it at target level with:

target_include_directories

CMakeLists.txt
```diff
@@ -59,85 +59,30 @@ if(BUILD_DECODER OR USE_ENCODER_SHADERC)
...
-# FFMPEG_FOUND - system has ffmpeg or libav
-# FFMPEG_INCLUDE_DIR - the ffmpeg include directory
-# FFMPEG_LIBRARIES
-# FFMPEG_LIBAVCODEC
-# FFMPEG_LIBAVFORMAT
-# FFMPEG_LIBAVUTIL
 if(FFMPEG_FOUND)
-    message("Found FFMPEG/LibAV libraries")
-    include_directories(${FFMPEG_INCLUDE_DIR})
+    message(STATUS "  libavcodec:  ${FFMPEG_LIBAVCODEC_LIBRARIES}")
```
dabrain34: use FFMPEG_LIBAVCODEC_LIBRARy

cmake/FindFFmpeg.cmake
```diff
@@ -1,230 +1,142 @@
...
+    PATH_SUFFIXES ffmpeg
+)
+
+# Common library search paths
+set(FFMPEG_LIB_SEARCH_PATHS
+    ${FFMPEG_ROOT}/lib
+    ${FFMPEG_WIN32_PREBUILT_DIR}/lib
+    ${PC_FFMPEG_LIBRARY_DIRS}
+    /Library/Frameworks
+    ~/Library/Frameworks
```
dabrain34: I dont think we gonna find any ffmpeg include dir here

dabrain34 (changes requested):

CMakeLists.txt
crmurillo: This will be addressed as part to PR #190 since the download logic is refactored there.

cmake/FindFFmpeg.cmake
crmurillo: Yes, path replaced with /opt/homebrew/

cmake/FindFFmpeg.cmake
crmurillo: Removed.

cmake/FindFFmpeg.cmake
crmurillo: Code refactor to make pkconfig have higher priority, however, if the user set the environment variable to use a custom installation, then FFMPEG_ROOT will be the priority.

PR #190, opened 2026-04-03
cmake: add logic to download ffmpeg for win32

crmurillo:
## Description

Currently, using FFmpeg on Windows requires manual installation.

This patch adds a cmake module that downloads FFmpeg shared libraries for Windows when they are not already present. Pre-built LGPL binaries are fetched from BtbN/FFmpeg-Builds for win64 and winarm64 platforms, and verified against published SHA256 checksums.

The automatic download can be disabled with `-DDOWNLOAD_FFMPEG=OFF`.

## Type of change

feature

### AMD Radeon RX 6600 (1002:73ff) / AMD proprietary driver 26.8.1 (AMD proprietary shader compiler) / Windows 10

Total Tests:    88
Passed:         37
Crashed:         7
Failed:          2
Not Supported:  42
Success Rate: 80.4%

### Intel(R) UHD Graphics 770 (ADL-S GT1) (8086:4680) / Intel open-source Mesa driver Mesa 26.3.0-devel (git-fe1491b758) / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         57
Crashed:         0
Failed:          0
Not Supported:  25
Success Rate: 100.0%

### AMD Radeon RX 7600 (RADV NAVI33) (1002:7480) / radv Mesa 26.3.0-devel (git-fe1491b758) / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  12
Success Rate: 100.0%

For AMD proprietary driver for Windows, 7 tests crash and 2 fail, same behavior as with the main branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -201,7 +201,7 @@ jobs:
   Vulkan-Video-Samples-win:
     strategy:
       matrix:
-        platform: [windows-x64, windows-x64-clang, windows-x64-ffmpeg, windows-x86, windows-x86-clang, win-arm64]
+        platform: [windows-x64, windows-x64-clang, windows-x64-without-ffmpeg, windows-x86, windows-x86-clang, win-arm64]
         include:
           - platform: windows-x64
             arch: x64
@@ -210,16 +210,19 @@ jobs:
             arch: x64
             cmake_arch: x64
             toolset: ClangCL
-          - platform: windows-x64-ffmpeg
+          - platform: windows-x64-without-ffmpeg
             arch: x64
             cmake_arch: x64
+            extra_cmake_flags: -DDOWNLOAD_FFMPEG=OFF
           - platform: windows-x86
             arch: x86
             cmake_arch: Win32
+            extra_cmake_flags: -DDOWNLOAD_FFMPEG=OFF
           - platform: windows-x86-clang
             arch: x86
             cmake_arch: Win32
             toolset: ClangCL
+            extra_cmake_flags: -DDOWNLOAD_FFMPEG=OFF
           - platform: win-arm64
             arch: arm64
             cmake_arch: ARM64
@@ -243,31 +246,6 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install FFmpeg
-        if: matrix.platform == 'windows-x64-ffmpeg'
-        run: |
-          # Define the custom installation location
-          $currentDir = pwd
-          $finalDestination = "$currentDir\vk_video_decoder\bin\libs\ffmpeg\win64"
-          echo $finalDestination
-
-          # Create the directory
-          New-Item -ItemType Directory -Force -Path $finalDestination
-
-          # Download FFmpeg static build
-          Invoke-WebRequest -Uri https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-master-latest-win64-gpl-shared.zip -OutFile ffmpeg.zip
-
-          # Extract FFmpeg to the tmp location
-          $tempExtractPath = "$currentDir\temp\extracted"
-
-          # Extract the entire archive to a temporary folder
-          Expand-Archive -Path ffmpeg.zip -DestinationPath $tempExtractPath -Force
-          Get-ChildItem -Path $tempExtractPath\ffmpeg-master-latest-win64-gpl-shared
-
-          # Get the contents of the first folder in the archive
-          Get-ChildItem -Path $tempExtractPath\ffmpeg-master-latest-win64-gpl-shared | Where-Object { $_.PSIsContainer } | Copy-Item -Destination $finalDestination -Force -Recurse
-          echo "$finalDestination\bin" | Out-File -Encoding ASCII -Append $env:GITHUB_PATH
-
       - name: Build Debug
         shell: bash
         run: |
@@ -275,6 +253,7 @@ jobs:
           cd BUILD_DEBUG
           cmake -DCMAKE_GENERATOR_PLATFORM=${{ matrix.cmake_arch }} \
             ${{ matrix.toolset && format('-T {0}', matrix.toolset) || '' }} \
+            ${{ matrix.extra_cmake_flags || '' }} \
             -DCMAKE_BUILD_TYPE=Debug ..
           cmake --build . --parallel $BUILD_JOBS --config Debug
 
@@ -285,6 +264,7 @@ jobs:
             pushd BUILD_RELEASE
             cmake -DCMAKE_GENERATOR_PLATFORM=${{ matrix.cmake_arch }} \
               ${{ matrix.toolset && format('-T {0}', matrix.toolset) || '' }} \
+              ${{ matrix.extra_cmake_flags || '' }} \
               -DCMAKE_BUILD_TYPE=Release ..
             cmake --build . --parallel $BUILD_JOBS --config Release
             mkdir release-artifacts
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -202,7 +202,7 @@ jobs:
     runs-on: ${{ matrix.os || 'windows-latest' }}
     strategy:
       matrix:
-        platform: [windows-x64, windows-x64-clang, windows-x64-ffmpeg, windows-x86, windows-x86-clang, win-arm64]
+        platform: [windows-x64, windows-x64-clang, windows-x64-without-ffmpeg, windows-x86, windows-x86-clang, win-arm64]
         include:
           - platform: windows-x64
             arch: x64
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -206,20 +206,25 @@ jobs:
           - platform: windows-x64
             arch: x64
             cmake_arch: x64
+            extra_cmake_flags: -DDOWNLOAD_FFMPEG=OFF
```
dabrain34: I would have a default value to ON but with an automatic detection in `cmake/FindFFmpeg.cmake`. I would not have separate files. I would download automatically the package in the right folder in case of non detection on Win32. The user will be able to pass `DOWNLOAD_FFMPEG=OFF` to cancel this auto download.

cmake/DownloadFFmpeg.cmake
```diff
@@ -0,0 +1,129 @@
+# DownloadFFmpeg.cmake
```
dabrain34: As said above I would not have a separate file to download ffmpeg. It should be included in FindFFMpeg.cmake

cmake/DownloadFFmpeg.cmake
```diff
@@ -0,0 +1,129 @@
...
+#
+# Automatically downloads pre-built FFmpeg shared libraries for Windows
+# from https://github.com/BtbN/FFmpeg-Builds/releases
+#
+# This module is only active on WIN32 builds when the FFmpeg libraries
```
dabrain34: I would detect that the module should be used only in Win32 but the call to the cmake module function should be in a Win32 branch

cmake/DownloadFFmpeg.cmake
```diff
@@ -0,0 +1,129 @@
...
+
+if((CMAKE_GENERATOR_PLATFORM MATCHES "^(aarch64|arm64|ARM64)"))
+    set(FFMPEG_ARCHIVE_NAME "ffmpeg-master-latest-winarm64-lgpl-shared")
+    set(FFMPEG_PLATFORM_DIR "winarm64")
+else()
+    set(FFMPEG_ARCHIVE_NAME "ffmpeg-master-latest-win64-lgpl-shared")
+    set(FFMPEG_PLATFORM_DIR "win64")
+endif()
+
+set(FFMPEG_DEST_DIR "${CMAKE_CURRENT_SOURCE_DIR}/vk_video_decoder/bin/libs/ffmpeg/${FFMPEG_PLATFORM_DIR}")
```
dabrain34: Please simplify using https://cmake.org/cmake/help/latest/module/FetchContent.html. 

You can decide a version to download from https://github.com/BtbN/FFmpeg-Builds/releases even if I would choose an official value such as 8.1 version and then pass the sha256 directly in the method `FetchContent_Declare`.

cmake/DownloadFFmpeg.cmake
```diff
@@ -0,0 +1,129 @@
...
+    if(FFMPEG_NUM_DIRS EQUAL 1)
+        list(GET FFMPEG_EXTRACTED_DIRS 0 FFMPEG_EXTRACTED_DIR)
+    else()
+        message(WARNING "Could not find extracted FFmpeg directory in ${FFMPEG_EXTRACT_DIR}")
+        return()
+    endif()
+endif()
+
+# Copy bin/, lib/, include/ into the destination
+foreach(SUBDIR bin lib include)
```
dabrain34: you would have to copy the content in the right folder from the FetchContent result directory

BUILD.md
```diff
@@ -61,17 +61,22 @@ Windows 10 or Windows 11 with the following software packages:
...
-  - Download the latest version of the FFMPEG shared libraries archive from https://github.com/BtbN/FFmpeg-Builds/releases
-  - The archive must have the following pattern in the name: For Windows x86_64 ffmpeg-*-win64-*-shared.zip
-  - Example download link:
-         For Windows x86_64 https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-master-latest-win64-gpl-shared.zip
-         For Windows ARM64  https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-master-latest-winarm64-lgpl-shared.zip
+- FFMPEG libraries (ONLY for Windows)
+  - FFmpeg is downloaded automatically during CMake configuration from https://github.com/BtbN/FFmpeg-Builds/releases
+  - The download is skipped if the libraries are already present in `vk_video_decoder\bin\libs\ffmpeg\<platform>\lib`
+  - To disable automatic download, pass `-DDOWNLOAD_FFMPEG=OFF` to CMake
+  - To download manually instead, get the latest version of the FFMPEG shared libraries archive from https://github.com/BtbN/FFmpeg-Builds/releases
```
dabrain34:
```suggestion
  - To download manually instead before the CMake configure, get the version of the FFMPEG shared libraries archive from any desired source.
```

BUILD.md
```diff
@@ -61,17 +61,22 @@ Windows 10 or Windows 11 with the following software packages:
...
+    - The archive must have the following pattern in the name:
```
dabrain34: Not really necessary

dabrain34 (changes requested):

.github/workflows/main.yml
crmurillo: yes, I'll remove the separate file.
The download happens by default, my idea here was to keep the previous behavior where we only had one job with ffmpeg enabled: `windows-x64-ffmpeg`.
@dabrain34: should we keep the same behavior or have all windows jobs download and build with ffmpeg now?

cmake/DownloadFFmpeg.cmake
crmurillo: Thanks for the hint, using `FetchContent` simplifies the logic a lot.

BUILD.md
crmurillo: I agree, line removed.

.github/workflows/main.yml
dabrain34: I would say that default value should be ON and one job should validate `DOWNLOAD_FFMPEG=OFF`

CMakeLists.txt
```diff
@@ -59,6 +59,10 @@ if(BUILD_DECODER OR USE_ENCODER_SHADERC)
     include(FindShaderc)
 endif()
 
+if(WIN32)
```
dabrain34: I think you can move this code in FindFFmpeg.cmake

.github/workflows/main.yml
crmurillo: All 64-bit jobs have now ffmpeg downloaded by default, with the exception of the new job `windows-x64-without-ffmpeg` that has `DOWNLOAD_FFMPEG=OFF`.
32-bit jobs have the download disabled by default as well since there are no ffmpeg binaries available.

dabrain34 (changes requested):

BUILD.md
```diff
@@ -61,17 +61,20 @@ Windows 10 or Windows 11 with the following software packages:
...
+  - The download is skipped if the libraries are already present in `vk_video_decoder\bin\libs\ffmpeg\<platform>\lib`
+  - To disable automatic download, pass `-DDOWNLOAD_FFMPEG=OFF` to CMake
+  - To download manually instead before the CMake configure, get the version of the FFMPEG shared libraries archive from any desired source.
+    - Example download link:
+         - For Windows x86_64 https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-master-latest-win64-lgpl-shared.zip
+         - For Windows ARM64  https://github.com/BtbN/FFmpeg-Builds/releases/download/latest/ffmpeg-master-latest-winarm64-lgpl-shared.zip
   - Extract to <APP_INSTALLED_LOC>\vk_video_decoder\bin\libs\ffmpeg
-  - Verify that <APP_INSTALLED_LOC>\vk_video_decoder\bin\libs\ffmpeg\win64\bin or <APP_INSTALLED_LOC>\vk_video_decoder\bin\libs\ffmpeg\winarm64\bin contains:
+  - Verify that <APP_INSTALLED_LOC>\vk_video_decoder\bin\libs\ffmpeg\win64\bin or <APP_INSTALLED_LOC>\vk_video_decoder\bin\libs\ffmpeg\winarm64\bin contains the required DLLs:
     - avformat-59.dll
```
dabrain34: NIT: can you update the revision to 62

dabrain34: you can also remote the file `download-ffmpeg-from.txt`

cmake/FindFFmpeg.cmake
```diff
@@ -7,6 +7,56 @@
 # FFMPEG_ROOT, if this module use this path to find FFMPEG headers
 # and libraries.
 
+if(WIN32)
+    option(DOWNLOAD_FFMPEG "Automatically download FFmpeg for Windows builds" ON)
+endif()
+
+if(WIN32 AND DOWNLOAD_FFMPEG)
+    set(FFMPEG_BTBN_TAG        "autobuild-2026-03-31-13-11")
```
dabrain34, line 15: I'm still wondering if it should passed through parameters as vulkan-headers to make it more generic. 

~And maybe have FindFFmpeg.cmake and DownloadFfmpeg.cmake.~

In the future it could be also great to make it available for Linux/Mac too

It leads also to the use DOWNLOAD_FFMPEG option, should it be guarded to WIN32 or should we support Linux/MacOS

dabrain34 (changes requested):

cmake/FindFFmpeg.cmake
crmurillo: Regarding passing the parameters, at the moment, I don't see any benefit. Vulkan-Headers is fetched as platform-independent source from a single git tag, so the version alone identifies it, but FFmpeg is fetched as per-platform prebuilt binaries verified by SHA256, so the pin is a (tag, revision, platform, hash) tuple.

I'd keep DOWNLOAD_FFMPEG guarded to WIN32 for now. Linux already gets FFmpeg from system packages via pkg-config in both BUILD.md and CI. Regarding macOS, I'm not sure what is the current support, at least,  there's no CI to test it.

PR #53, opened 2025-06-03
Backport:  integrate vk video samples from NVPro workspace, part 2

dabrain34:
## Description

Backport commits from vk_video_samples to align the repositories with fixes found in NVPro space

Warning: There was a regression on both encode and decode with 10 bits media.

## Type of change

Backport

### Intel(R) Iris(R) Xe Graphics (ADL GT2) / Intel open-source Mesa driver Mesa 26.1.0-devel (git-6dce207497) / Ubuntu 24.04.4 LTS

Total Tests:    70
Passed:         50
Crashed:         0
Failed:          0
Not Supported:   6
Skipped:        14 (in skip list)
Success Rate: 100.0%

### NVIDIA GeForce RTX 3050 Ti Laptop GPU / NVIDIA 580.94.15 / Ubuntu 24.04.4 LTS

Total Tests:    70
Passed:         56
Crashed:         0
Failed:          0
Not Supported:   7
Skipped:         7 (in skip list)
Success Rate: 100.0%

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.1.0-devel (git-fcd280f286) / Ubuntu 24.04.4 LTS

Total Tests:    73
Passed:         62
Crashed:         0
Failed:          0
Not Supported:   6
Skipped:         5 (in skip list)
Success Rate: 100.0%

### LAST COMMIT
001c2ff Fix frame to file from adding .yuv to filenames with .y4m already

### BACKPORT
5376ce2 Add cast-qual to the cmake settings
001c2ff Fix frame to file from adding .yuv to filenames with .y4m already
16607fe encoder: Deal with the Vulkan chained stuctures
441343e common: Use the CRC generator instead of embedding it
88875e4 common: Code compilation fixes (modified as there was a regression on 10 bits)
db43c78 encode: remove the CPU input conversion function
593777c Modify the output filename extension based on the type

### OPTIONAL:
77ca65c github actions: Build Windows only with FFMPEG
98a13b0 github actions: add a job to build with ffmpeg
1b1df8b Add option to generate raw YUV frames
b9df139 Instead of defining the destructor inline, provide its implementation in a separate source file to make GCC 13 happy in release mode. Also fix other compilation errors in release mode.
624a0ad Add Duck IVF parsing support to the elementary stream.
1f5ebf0 decode: make the Y4M file format to be the default output

### BACKPORTED:
f6fdb5e Remove duplicate messages about finding shaderc
a0497b5 Remove unused variables and code related to shaderc
[...]

PR #194, opened 2026-04-20
encode: fixes related to pNext-chaining

srinathkr-nv:
## Description

This change fixes a regression arising from https://github.com/KhronosGroup/Vulkan-Video-Samples/commit/62c184cc00879137217c01c1575c9c280e147df4 which prevented the quantization maps functionality from working correctly and also regressed RC struct chaining code.

## Type of change

Bug fix (and associated minor cleanup)

## Issue (optional)

See https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/53#issuecomment-4177777784

### NVIDIA L2 / NVIDIA 595.44.05 / Ubuntu 24.04.4 LTS

Total Tests:    73
Passed:         62
Crashed:         0
Failed:          0
Not Supported:   9
Skipped:         2 (in skip list)
Success Rate: 100.0%

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
@@ -1930,13 +1930,10 @@ VkResult VkVideoEncoder::RecordVideoCodingCmd(VkSharedBaseObj<VkVideoEncodeFrame
         videoInlineQueryInfoKHR.queryPool = queryPool;
         videoInlineQueryInfoKHR.firstQuery = querySlotId;
         videoInlineQueryInfoKHR.queryCount = numQuerySamples;
-        VkBaseInStructure* pStruct = (VkBaseInStructure*)&encodeFrameInfo->encodeInfo;
-        vk::ChainNextVkStruct(*pStruct, videoInlineQueryInfoKHR);
+
+        vk::ChainNextVkStruct(encodeFrameInfo->encodeInfo, videoInlineQueryInfoKHR);
 
         vkDevCtx->CmdEncodeVideoKHR(cmdBuf, &encodeFrameInfo->encodeInfo);
-
-        // Remove the stack pointer from the chain, causes a use after free otherwise in GetEncodeFrameInfoH264
-        encodeFrameInfo->encodeInfo.pNext = videoInlineQueryInfoKHR.pNext;
     }
     else
     {
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
@@ -1916,9 +1916,10 @@ VkResult VkVideoEncoder::RecordVideoCodingCmd(VkSharedBaseObj<VkVideoEncodeFrame
                                                           encodeFrameInfo->controlCmd};
         vkDevCtx->CmdControlVideoCodingKHR(cmdBuf, &renderControlInfo);
 
+        assert(encodeFrameInfo->pControlCmdChain);
+
         m_beginRateControlInfo = *(VkVideoEncodeRateControlInfoKHR*)encodeFrameInfo->pControlCmdChain;
-        // Do not walk the chain, otherwise we end up creating a loop here.
-        m_beginRateControlInfo.pNext = NULL;
+        const_cast<VkBaseInStructure*>(static_cast<const VkBaseInStructure*>(m_beginRateControlInfo.pNext))->pNext = NULL;
     }
 
     if (m_videoMaintenance1FeaturesSupported)
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
@@ -56,10 +56,7 @@ public:
 
     struct VkVideoEncodeFrameInfo : public VkVideoRefCountBase
     {
-        VkStructureType GetType() {
-            return (encodeInfo.pNext == nullptr) ?
-                    VK_STRUCTURE_TYPE_VIDEO_ENCODE_INFO_KHR : reinterpret_cast<const VkBaseInStructure*>(encodeInfo.pNext)->sType;
-        }
+        virtual VkVideoCodecOperationFlagBitsKHR GetCodecType() const = 0;
 
         VkVideoEncodeFrameInfo(const void* pNext = nullptr)
             : encodeInfo{ VK_STRUCTURE_TYPE_VIDEO_ENCODE_INFO_KHR, pNext}
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
@@ -250,7 +250,8 @@ public:
         virtual void Reset(bool releaseResources = true) {
             // Clear and check state
             assert(encodeInfo.sType == VK_STRUCTURE_TYPE_VIDEO_ENCODE_INFO_KHR);
-            assert(encodeInfo.pNext != nullptr);
+
+            encodeInfo.pNext = nullptr;
 
             if ((frameInputOrderNum == (uint64_t)-1) &&
                 (frameEncodeInputOrderNum == (uint64_t)-1) &&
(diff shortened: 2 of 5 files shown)


srinathkr-nv:
@dabrain34 @lolzballs please review these changes.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
```diff
@@ -56,12 +56,12 @@ class VkVideoEncoder : public VkVideoRefCountBase {
 
     struct VkVideoEncodeFrameInfo : public VkVideoRefCountBase
     {
-        VkStructureType GetType() {
-            return (encodeInfo.pNext == nullptr) ?
-                    VK_STRUCTURE_TYPE_VIDEO_ENCODE_INFO_KHR : reinterpret_cast<const VkBaseInStructure*>(encodeInfo.pNext)->sType;
+        inline VkVideoCodecOperationFlagBitsKHR GetType() const {
```
dabrain34: in c++17 a class member function is already an inline function so no need to redefine it. This is redundant.

I would  prefer to rename the mehod `GetCodecType` as in `vkVideoCoreProfile.h`
Dont forget to print an error in release. This assert is valid only in Debug and this might lead to unknown behavior

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
dabrain34, line 1812: this is a dead branch as pNext will be always null, see +1794

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1917,8 +1917,7 @@ VkResult VkVideoEncoder::RecordVideoCodingCmd(VkSharedBaseObj<VkVideoEncodeFrame
         vkDevCtx->CmdControlVideoCodingKHR(cmdBuf, &renderControlInfo);
 
         m_beginRateControlInfo = *(VkVideoEncodeRateControlInfoKHR*)encodeFrameInfo->pControlCmdChain;
```
dabrain34, line 1921: I think here i would search for the RateControlInfo object in the chaining of VkStruct and then affect m_beginRateControlInfo to this object found or create a new one if it does not exist

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.h
```diff
@@ -75,6 +75,9 @@ class VkVideoEncoderAV1 : public VkVideoEncoder
 
             // Reset the base first
             VkVideoEncodeFrameInfo::Reset(releaseResources);
+
+            // After resetting the base structure parameters, start building the pNext chain again
+            encodeInfo.pNext = &pictureInfo;
```
dabrain34, line 85: I would check state as in VkVideoEncodeFrameInfoH264::Reset such as 

```
 assert(pictureInfo.sType == VK_STRUCTURE_TYPE_VIDEO_ENCODE_AV1_PICTURE_INFO_KHR);
...
```

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.h
```diff
@@ -184,7 +187,7 @@ class VkVideoEncoderAV1 : public VkVideoEncoder
 
 private:
     VkVideoEncodeFrameInfoAV1* GetEncodeFrameInfoAV1(VkSharedBaseObj<VkVideoEncodeFrameInfo>& encodeFrameInfo) {
-        assert(VK_STRUCTURE_TYPE_VIDEO_ENCODE_AV1_PICTURE_INFO_KHR == encodeFrameInfo->GetType());
+        assert(VK_VIDEO_CODEC_OPERATION_ENCODE_AV1_BIT_KHR == encodeFrameInfo->GetType());
```
dabrain34: use GetCodecType

dabrain34 (changes requested):

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
```diff
@@ -358,6 +360,7 @@ class VkVideoEncoder : public VkVideoRefCountBase {
         std::atomic<int32_t>                m_refCount;
         VkSharedBaseObj<VulkanBufferPoolIf> m_parent;
         int32_t                             m_parentIndex;
```
lolzballs, line 358: Is there an advantage to making this a member variable? It would be cleaner to make `GetType()` pure virtual and have the codec-specific classes override it (or even use RTTI to avoid any code to drive this).

`GetType()` is only called for some sanity-checks anyways, so I don't think the overhead of such calls is a valid concern. (ideally we could add NDEBUG to release builds).

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
lolzballs: I think worrying about release is unnecessary. To me this assertion is a sanity check which should only fail if the developer is doing something weird (and the developer is probably not using release mode with NDEBUG).

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
dabrain34: I agree that a pure virtual + codec specific override looks more elegant

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
srinathkr-nv: I know, but I'm not planning on touching these lines in this PR because the focus is on addressing issues caused by commit 62c184cc. I have yet another set of changes in progress for cleaning up various pNext-chaining issues in the codebase.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
srinathkr-nv: Same as in my other response, I have yet another set of changes in progress for cleaning up various pNext-chaining issues in the codebase and I'll tackle this as part of that set.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.h
srinathkr-nv: My updated set of commits addresses this feedback.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.h
srinathkr-nv: I have addressed this in my updated set of commits.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
dabrain34: At least I would validate in this patch that pControlCmdChain is a valid pointer different from nullptr even this is unlikely to happen.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
dabrain34: ok lets go for another patch in another PR

dabrain34:
can you add a guard as mention in the thread,  rebase, fix the buildbot with reset  ?

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
dabrain34: maybe good to have an assert here if its unlikely to happen

lolzballs (approved):

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
srinathkr-nv: Ok, I have added an assert in my updated set of changes now.

srinathkr-nv:
> can you add a guard as mention in the thread, rebase, fix the buildbot with reset ?

I don't know why the build bot decided to complain about the `Reset` methods for the changes in this PR as that is a preexisting issue, but I have added a commit to address that too now.

dabrain34 (approved):

PR #197, opened 2026-04-23
tests: automate sample list addition with pre filled entries such as filepath, sha256 and md5

dabrain34:
## Description

Fill up fields automatically such as file_path, sha256 or md5 decode result when adding new samples to the decode and encode sample list.

## Type of change

 feature

### NVIDIA GeForce RTX 3050 Ti Laptop GPU / NVIDIA 595.44.00 / Ubuntu 24.04.4 LTS

Total Tests:    73
Passed:         58
Crashed:         0
Failed:          0
Not Supported:  14
Success Rate: 100.0%


crmurillo (approved):
lgtm, thanks!

PR #201, opened 2026-05-06
tests: clean up the results and provide csv output similar to fluster

dabrain34:
## Description

The report is now unified and similar to fluster result symbols. It is possible to convert json to csv

## Type of change

cleanup

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.0.6 (git-0e095aab43) / Ubuntu 24.04.4 LTS

Total Tests:    73
Passed:         63
Crashed:         0
Failed:          0
Not Supported:   6
Success Rate: 100.0%

### DIFF
--- a/tests/libs/video_test_result_reporter.py
+++ b/tests/libs/video_test_result_reporter.py
@@ -31,14 +31,30 @@ def get_status_display(status: VideoTestStatus) -> tuple:
         Tuple of (status_text, status_symbol)
     """
     if status == VideoTestStatus.SUCCESS:
-        return "PASS", "✓"
+        return "PASS", "✔️"
     if status == VideoTestStatus.NOT_SUPPORTED:
         return "N/S", "○"
     if status == VideoTestStatus.CRASH:
         return "CRASH", "💥"
     if status == VideoTestStatus.SKIPPED:
         return "SKIP", "⊘"
-    return "FAIL", "✗"
+    return "FAIL", "❌"
+
+
+def get_status_symbol_from_str(status) -> str:
+    """Return the symbol for a status, accepting either a status string
+    (e.g. "success") or a VideoTestStatus enum value.
+    """
+    if isinstance(status, VideoTestStatus):
+        status = status.value
+    symbols = {
+        VideoTestStatus.SUCCESS.value: "✔️",
+        VideoTestStatus.NOT_SUPPORTED.value: "○",
+        VideoTestStatus.CRASH.value: "💥",
+        VideoTestStatus.SKIPPED.value: "⊘",
+        VideoTestStatus.ERROR.value: "❌",
+    }
+    return symbols.get(status, status)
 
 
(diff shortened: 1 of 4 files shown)


tests/libs/video_test_result_reporter.py
```diff
@@ -31,14 +31,36 @@ def get_status_display(status: VideoTestStatus) -> tuple:
...
         return "N/S", "○"
     if status == VideoTestStatus.CRASH:
         return "CRASH", "💥"
     if status == VideoTestStatus.SKIPPED:
         return "SKIP", "⊘"
-    return "FAIL", "✗"
+    return "FAIL", "❌"
+
+
+def get_status_symbol_from_str(status) -> tuple:
```
crmurillo: The return type is str not a tuple.

tests/libs/video_test_result_reporter.py
```diff
@@ -31,14 +31,36 @@ def get_status_display(status: VideoTestStatus) -> tuple:
...
+    Returns:
+        status_symbol
+    """
+    if status == "success":
+        return "✔️"
+    if status == "not_supported":
+        return "○"
+    if status == "crash":
+        return "💥"
+    if status == "skip":
```
crmurillo: status should be compared against "skipped"

tests/vvs_test_runner.py
```diff
@@ -520,6 +517,40 @@ def export_results_json(self, output_file: str) -> bool:
             print(f"✗ Failed to export results: {e}")
             return False
 
+    def export_results_csv(self, output_file: str) -> bool:
+        """Export test results to JSON file"""
```
crmurillo:
```suggestion
        """Export test results to CSV file"""
```

tests/libs/video_test_result_reporter.py
```diff
@@ -31,14 +31,36 @@ def get_status_display(status: VideoTestStatus) -> tuple:
...
+    """Get status display string and symbol
+
+    Args:
+        status: status string
+
+    Returns:
+        status_symbol
+    """
+    if status == "success":
+        return "✔️"
```
crmurillo: Should the other symbols be updated as well? At least those relevant to the test results, since it currently looks like this:
```
[4/21] Running: encode_h264_high_profile
  🔍 Validating with decoder: test_output_h264_high_profile.264
  ✓ Decoder validation passed
✔️ PASS (0.35s)
```

```
[19/53] Running: decode_av1_globalmotion_8bit
✓ MD5 verification passed: 74f9f9bbca22da5ffdaee2833f7bc1c9
✔️ PASS (0.11s)

```

dabrain34: it was more like keeping the same as fluster for the result. So I would prefer to keep the ✔️ for the test result only

crmurillo (approved):
LGTM, thanks for the patch!

PR #210, opened 2026-05-31
tests: add --validate to enable the Vulkan validation layer

dabrain34:
## Description

When --validate is passed, export
VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation
into each test subprocess environment so the validation layer is loaded. When the layer reports messages (VUID-/Validation Error/Warning), surface the captured output even on PASS so validation issues are not silently swallowed.

## Type of change

feature

### NVIDIA GeForce RTX 3050 Ti Laptop GPU / NVIDIA 595.44.00 / Ubuntu 24.04.4 LTS

Total Tests:    79
Passed:         62
Crashed:         0
Failed:          0
Not Supported:  16
Success Rate: 100.0%

### DIFF
--- a/tests/libs/video_test_framework_base.py
+++ b/tests/libs/video_test_framework_base.py
@@ -19,6 +19,7 @@ limitations under the License.
 
 import fnmatch
 import json
+import os
 import shutil
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


crmurillo (approved):
LGTM, thanks for the patch!

PR #219, opened 2026-06-23
encoder: build the static lib with the shaderc compute filter

crmurillo:
## Description

When the encoder is consumed as a static library by a project that embeds vk_video_encoder/libs via add_subdirectory(), the shaderc GPU compute filter is silently disabled even when shaderc is available. 

At runtime the encoder throws the following warning:
```
Warning: input format (...) differs from encode format (...) but compute filter
is not available (built without SHADERC_SUPPORT). Output may be incorrect.
```
There are two issues:
- `USE_ENCODER_SHADERC` is only declared in the top-level build. Consumers that add the libs directory directly never see the option, so it stays undefined and the filter is off, regardless of whether shaderc was found.
- The static target never receives `SHADERC_SUPPORT`. Only the shared target gets the definition, so for the static target, VkVideoEncoder.cpp compiles the no-filter path even when `USE_ENCODER_SHADERC` is on.

## Type of change

bug fix

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.1.0 (git-8d005e3d66) / Ubuntu 24.04.4 LTS

Total Tests:    83
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%

### DIFF
--- a/vk_video_encoder/libs/CMakeLists.txt
+++ b/vk_video_encoder/libs/CMakeLists.txt
@@ -84,9 +84,24 @@ set(LIBVKVIDEOENCODER_SRC
     ${VK_VIDEO_COMMON_LIBS_SOURCE_ROOT}/VkCodecUtils/VulkanBistreamBufferImpl.cpp
 )
 
-set(LIBVKVIDEOENCODER_DEFINITIONS
+# Common defines applied to both the shared and static libraries
+set(LIBVKVIDEOENCODER_DEFINITIONS "")
+
+# Defines specific to the shared library
+set(LIBVKVIDEOENCODER_SHARED_DEFINITIONS
     PRIVATE VK_VIDEO_ENCODER_IMPLEMENTATION
     PUBLIC VK_VIDEO_ENCODER_SHAREDLIB)
+
+# When this directory is added directly via add_subdirectory() by a parent that does
+# not declare the USE_ENCODER_SHADERC option, default it from shaderc availability so the
+# GPU compute filter is enabled whenever shaderc was found
+if(NOT DEFINED USE_ENCODER_SHADERC)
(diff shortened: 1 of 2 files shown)


vk_video_encoder/libs/CMakeLists.txt
```diff
@@ -140,6 +151,7 @@ add_library(${VULKAN_VIDEO_ENCODER_STATIC_LIB} STATIC ${LIBVKVIDEOENCODER_SRC})
 if(USE_ENCODER_SHADERC)
     # Link the libraries
     target_link_libraries(${VULKAN_VIDEO_ENCODER_STATIC_LIB} PUBLIC ${SHADERC_SHARED_LIBRARY})
+    target_compile_definitions(${VULKAN_VIDEO_ENCODER_STATIC_LIB} PRIVATE SHADERC_SUPPORT)
```
dabrain34: I would use $(LIBVKVIDEOENCODER_DEFINITIONS) and put it outside of the if block

```suggestion
    target_compile_definitions(${VULKAN_VIDEO_ENCODER_STATIC_LIB} PRIVATE $(LIBVKVIDEOENCODER_DEFINITIONS))
```

As it, this is bringing an issue as it will define `VK_VIDEO_ENCODER_SHAREDLIB` in static library build definitions so I would recommend to rename `LIBVKVIDEOENCODER_DEFINITIONS` to `LIBVKVIDEOENCODER_SHARED_DEFINITIONS` with the shared related defines and define a common one `LIBVKVIDEOENCODER_DEFINITIONS` with the SHADERC_SUPPORT for both shared and static libraries.

dabrain34 (changes requested):

PR #222, opened 2026-07-04
encoder: default msbShift for >8-bit input

crmurillo:
## Description

The compute-filter input path only shifts 10/12-bit samples into the most-significant bits when input.msbShift > 0, but the default is -1. So without an explicit --msbShift, high bit-depth input reached the encoder LSB-aligned and encoded to garbage. 

This has side-effects as well if the encoder library is used by other application like Vulkan CTS. If the application does not pass --msbShift explicitly, the result is also corrupted.

## Type of change

bug fix

Total Tests:    86
Passed:         71
Crashed:         0
Failed:          0
Not Supported:  11
Success Rate: 100.0%

The issue is reproducible running the following command:
`python3 vvs_test_runner.py -t av1_main_profile_10bit --keep-files`

This will create the ivf file in: 
`$VVS_DIR/tests/results/test_output_av1_main_profile_10bit.ivf`

The generated video `test_output_av1_main_profile_10bit.ivf` can be played back using ffplay or another application. The result is a green video.

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
@@ -663,6 +663,10 @@ int EncoderConfig::ParseArguments(int argc, const char *argv[])
         return -1;
     }
 
+    if (input.msbShift < 0 && input.bpp > 8) {
+        input.msbShift = static_cast<int8_t>(16 - input.bpp);
+    }
+
     frameCount = inputFileHandler.GetFrameCount(input.width, input.height, input.bpp, input.chromaSubsampling);
 
     if (numFrames == 0 || numFrames > frameCount) {
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
@@ -217,12 +217,9 @@ VkResult VkVideoEncoder::LoadNextFrame(VkSharedBaseObj<VkVideoEncodeFrameInfo>& 
                         width, height);
             }
         } else if (m_encoderConfig->input.bpp == 10 || m_encoderConfig->input.bpp == 12) {
-            int shiftBits = 0;
-            if (m_encoderConfig->input.msbShift >= 0) {
-                shiftBits = m_encoderConfig->input.msbShift;
-            } else {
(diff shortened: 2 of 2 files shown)


tests/encode_samples.json
```diff
@@ -172,7 +172,9 @@
       "profile": "main10",
       "extra_args": [
         "--inputBpp",
-        "10"
```
dabrain34, line 175: I would remove the msbShift parameter as it seems useless and should be calculated according to the inputBpp internally.

srinathkr-nv: I'm confused a bit by this comment, and the change made to VkEncoderConfig.cpp. With the change in VkEncoderConfig.cpp, we should not need to specify the MSB shift explicitly, correct?

@dabrain34 the `msbShift` option should not be removed, because it is helpful for disabling MSB shifting if the input is already in the `P010` format, for example. Or, did you mean that we should not need to specify the MSB shift in the test configuration here (so, identical to what I said above)?

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -663,6 +663,10 @@ int EncoderConfig::ParseArguments(int argc, const char *argv[])
         return -1;
     }
 
+    if (input.msbShift < 0 && input.bpp > 8) {
+        input.msbShift = static_cast<int8_t>(16 - input.bpp);
```
srinathkr-nv, line 667: With this change, we are ensuring that `input.msbShift` will have a non-negative value going forward. This can be an opportunity to simplify the following code block in VkVideoEncoder.cpp, around line 217:

```C++
    int shiftBits = 0;
    if (m_encoderConfig->input.msbShift >= 0) {
        shiftBits = m_encoderConfig->input.msbShift;
    } else {
        shiftBits = 16 - m_encoderConfig->input.bpp;
    }
```

crmurillo: agreed, will simplify this as well

dabrain34: I agree and forgot about this use case. Thanks @srinathkr-nv

dabrain34:
LGTM @srinathkr-nv  ok for you ?

srinathkr-nv (approved):
I got around to testing this out just now; LGTM too.

PR #224, opened 2026-07-13
VkVideoEncoder: add VK_KHR_video_encode_feedback2 support for AV1

elima:
## Description

Adds encoder support for the `VK_KHR_video_encode_feedback2` extension to the AV1 encode path, exposing the richer per-encode feedback the extension provides on top of the base `VK_KHR_video_encode_queue` feedback.

When any feedback2 option is enabled, the encoder now requests the `VK_KHR_video_encode_feedback2` device extension and the `videoEncodeFeedback2` feature, sets up the appropriate query pool, and parses the returned feedback into:

- **Encode status**, bitstream **size** and **offset**
- **Picture-level feedback**: average / min / max quantization
- **Pixel counts**: intra / inter / skipped
- **Picture partition count**
- **Per-partition feedback**: status, bitstream buffer offset and bytes written, up to a requested maximum number of entries

These outputs are intended for use by VK-GL-CTS to validate encoded output against the decoded bitstream, in particular for the per-partition count tests.

### New encoder CLI options

| Option | Description |
| --- | --- |
| `--pictureFeedback` | Enable picture-level feedback2 output |
| `--params` | Provide custom picture feedback params (used together with `--pictureFeedback`; otherwise defaults are used) |
| `--pixelFeedback` | Enable pixel-count feedback2 output (intra/inter) |
| `--skippedPixelFeedback` | Enable skipped-pixel feedback2 output (implies `--pixelFeedback`) |
| `--enablePerPartitionFeedback` | Enable per-partition feedback2 output |
| `--maxPerPartitionFeedbackEntries <n>` | Max per-partition feedback entries to request |

### Build requirement

`VK_KHR_video_encode_feedback2` first appears in **Vulkan-Headers v1.4.353**, so the minimum required patch version is bumped from 321 to 353. When the system SDK is older, the build fetches v1.4.353 automatically. `VkVSCommon.h` now routes Vulkan through `vulkan_interfaces.h` so the encoder library also builds under VKCTS without relying on a system Vulkan installation.

## Type of change

feature

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.1.0-devel (git-16ece32984) / Debian GNU/Linux 13 (trixie)

Total Tests:    83
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%

### NVIDIA GeForce RTX 4060 Ti / NVIDIA 595.44.00 / Debian GNU/Linux 13 (trixie)

Total Tests:    83
Passed:         72
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%

### DIFF
--- a/common/libs/VkCodecUtils/VulkanDeviceContext.cpp
+++ b/common/libs/VkCodecUtils/VulkanDeviceContext.cpp
@@ -797,6 +797,11 @@ VkResult VulkanDeviceContext::CreateVulkanDevice(int32_t numDecodeQueues,
                                                                           false // videoEncodeAV1
                                                                         };
 
+        VkPhysicalDeviceVideoEncodeFeedback2FeaturesKHR feedback2Features { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VIDEO_ENCODE_FEEDBACK_2_FEATURES_KHR,
+                                                                            nullptr,
+                                                                            false // videoEncodeFeedback2
+                                                                          };
+
         // Chain only the structures that are requested
         VkBaseInStructure* pNext = nullptr;
         if (videoCodecs & VK_VIDEO_CODEC_OPERATION_ENCODE_AV1_BIT_KHR) {
@@ -806,6 +811,10 @@ VkResult VulkanDeviceContext::CreateVulkanDevice(int32_t numDecodeQueues,
         if (videoCodecs & VK_VIDEO_CODEC_OPERATION_DECODE_VP9_BIT_KHR) {
             videoDecodeVP9Feature.pNext = pNext;
             pNext = (VkBaseInStructure*)&videoDecodeVP9Feature;
+        }
+        if (m_videoEncodeFeedback2Enabled) {
+            feedback2Features.pNext = pNext;
+            pNext = (VkBaseInStructure*)&feedback2Features;
         }
 
         VkPhysicalDeviceTimelineSemaphoreFeatures timelineSemaphoreFeatures { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TIMELINE_SEMAPHORE_FEATURES,
@@ -844,6 +853,8 @@ VkResult VulkanDeviceContext::CreateVulkanDevice(int32_t numDecodeQueues,
                              (videoEncodeAV1Feature.videoEncodeAV1 != VK_FALSE), "videoEncodeAV1", false);
         CHECK_VULKAN_FEATURE(((videoCodecs & VK_VIDEO_CODEC_OPERATION_DECODE_VP9_BIT_KHR) != 0) ==
                              (videoDecodeVP9Feature.videoDecodeVP9 != VK_FALSE), "videoDecodeVP9", false);
+        CHECK_VULKAN_FEATURE((m_videoEncodeFeedback2Enabled != 0) ==
+                             (feedback2Features.videoEncodeFeedback2 != VK_FALSE), "videoEncodeFeedback2", false);
 
         // Validate feature support here.
         // TODO: Currntly this method is receiving all codec bits irrespective of the codec that is required to decode/encode provided input.
@@ -974,6 +985,7 @@ VulkanDeviceContext::VulkanDeviceContext()
     , m_importedDeviceHandle(false)
     , m_videoDecodeQueryResultStatusSupport(false)
     , m_videoEncodeQueryResultStatusSupport(false)
+    , m_videoEncodeFeedback2Enabled(false)
     , m_device()
     , m_gfxQueue()
     , m_computeQueue()
--- a/common/libs/VkCodecUtils/VulkanDeviceContext.h
+++ b/common/libs/VkCodecUtils/VulkanDeviceContext.h
@@ -111,6 +111,13 @@ public:
     bool    GetVideoEncodeQueryResultStatusSupport() const { return m_videoEncodeQueryResultStatusSupport; }
     VkQueueFlags GetVideoDecodeQueueFlag() const { return m_videoDecodeQueueFlags; }
     VkQueueFlags GetVideoEncodeQueueFlag() const { return m_videoEncodeQueueFlags; }
+    void SetVideoEncodeFeedback2Enabled(bool enable) {
+        m_videoEncodeFeedback2Enabled = enable;
+        if (enable) {
+            AddReqDeviceExtension(VK_KHR_VIDEO_ENCODE_FEEDBACK_2_EXTENSION_NAME);
+        }
+    }
+    bool GetVideoEncodeFeedback2Enabled() const { return m_videoEncodeFeedback2Enabled; }
     class MtQueueMutex {
 
     public:
@@ -341,6 +348,7 @@ private:
     uint32_t m_importedDeviceHandle : 1;
     uint32_t m_videoDecodeQueryResultStatusSupport : 1;
     uint32_t m_videoEncodeQueryResultStatusSupport : 1;
+    uint32_t m_videoEncodeFeedback2Enabled : 1;
     VkDevice                m_device;
     VkQueue                 m_gfxQueue;
     VkQueue                 m_computeQueue;
--- a/tests/encode_samples.json
+++ b/tests/encode_samples.json
@@ -277,6 +277,39 @@
       "source_filepath": "video/yuv/352x288_15_i420.yuv"
     },
     {
+      "name": "av1_encode_feedback",
+      "codec": "av1",
+      "profile": "main",
+      "extra_args": [
+        "--pictureFeedback",
+        "--pixelCountFeedback",
+        "--skippedPixelCountFeedback"
+      ],
+      "description": "Test AV1 picture-level encode feedback2 (qp, pixel counts)",
+      "width": 1920,
+      "height": 1080,
+      "source_url": "https://storage.googleapis.com/vulkan-video-samples/yuv/1920x1080_420_8le.yuv",
+      "source_checksum": "9cdc96522bd15977445459aa144d019029e65e76c1893c11f89666645e8afc1e",
+      "source_filepath": "video/yuv/1920x1080_420_8le.yuv"
+    },
+    {
+      "name": "av1_encode_feedback_per_partition",
+      "codec": "av1",
+      "profile": "main",
+      "extra_args": [
+        "--pictureFeedback",
+        "--enablePerPartitionFeedback",
+        "--maxPerPartitionFeedbackEntries",
+        "16"
+      ],
+      "description": "Test AV1 per-partition encode feedback2",
+      "width": 1920,
+      "height": 1080,
+      "source_url": "https://storage.googleapis.com/vulkan-video-samples/yuv/1920x1080_420_8le.yuv",
+      "source_checksum": "9cdc96522bd15977445459aa144d019029e65e76c1893c11f89666645e8afc1e",
+      "source_filepath": "video/yuv/1920x1080_420_8le.yuv"
+    },
+    {
       "name": "h264_resolution_too_small",
       "codec": "h264",
       "profile": null,
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
@@ -144,6 +144,11 @@ static void printHelp(VkVideoCodecOperationFlagBitsKHR codec)
                                         Eg: 1. \"--cdef --params 3  2  1 2 9 1   2 3 7 3   3 1 4 2   4 0 3 2\"\n\
                                             2. \"--cdef --params 3  0  1 2 9 1\"\n\
                                             3. \"--cdef\"\n\
+        --pictureFeedback               Enable picture-level feedback2 output\n\
+        --pixelCountFeedback            Enable pixel count feedback2 output (intra/inter)\n\
+        --skippedPixelCountFeedback     Enable skipped pixel count feedback2 output (implies --pixelCountFeedback)\n\
+        --enablePerPartitionFeedback    Enable per-partition feedback2 output\n\
+        --maxPerPartitionFeedbackEntries <integer> : Max per-partition feedback entries to request\n\
         --lr                            Enable loop restoration filter\n\
         --params                        Enabel custom loop restoration filter configuration when followed by --lr option\n\
                                         Otherwise default loop restoration filter configuration will be used\n\
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
@@ -139,6 +139,17 @@ int EncoderConfigAV1::DoParseArguments(int argc, const char* argv[])
                     READ_PARAM(i, lrConfig.LoopRestorationSize[j], uint16_t);
                 }
             }
+        } else if (args[i] == "--pictureFeedback") {
+            enablePictureFeedback = true;
+        } else if (args[i] == "--pixelCountFeedback") {
+            enablePixelCountFeedback = true;
+        } else if (args[i] == "--skippedPixelCountFeedback") {
+            enableSkippedPixelCountFeedback = true;
+            enablePixelCountFeedback = true;
+        } else if (args[i] == "--enablePerPartitionFeedback") {
+            enablePerPartitionFeedback = true;
+        } else if (args[i] == "--maxPerPartitionFeedbackEntries") {
+            READ_PARAM(i, maxPerPartitionFeedbackEntries, uint32_t);
         } else if (args[i] == "--profile"){
             if (++i >= argc) {
                 fprintf(stderr, "invalid parameter for %s\n", args[i-1].c_str());
@@ -190,6 +201,12 @@ bool EncoderConfigAV1::InitSequenceHeader(StdVideoAV1SequenceHeader *seqHdr,
 
 VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    const bool feedback2Requested = enablePictureFeedback || enablePixelCountFeedback ||
+                                    enableSkippedPixelCountFeedback || enablePerPartitionFeedback;
+    videoEncodeFeedback2Capabilities = VkVideoEncodeFeedback2CapabilitiesKHR
+        { VK_STRUCTURE_TYPE_VIDEO_ENCODE_FEEDBACK_2_CAPABILITIES_KHR, nullptr };
+    void* pExtCapabilities = feedback2Requested ? &videoEncodeFeedback2Capabilities : nullptr;
+
     VkResult result = VulkanVideoCapabilities::GetVideoEncodeCapabilities<VkVideoEncodeAV1CapabilitiesKHR, VK_STRUCTURE_TYPE_VIDEO_ENCODE_AV1_CAPABILITIES_KHR,
                                                                           VkVideoEncodeAV1QuantizationMapCapabilitiesKHR, VK_STRUCTURE_TYPE_VIDEO_ENCODE_AV1_QUANTIZATION_MAP_CAPABILITIES_KHR>
                                                         (vkDevCtx, videoCoreProfile,
@@ -198,7 +215,8 @@ VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkD
                                                          av1EncodeCapabilities,
                                                          quantizationMapCapabilities,
                                                          av1QuantizationMapCapabilities,
-                                                         intraRefreshCapabilities);
+                                                         intraRefreshCapabilities,
+                                                         pExtCapabilities);
     if (result != VK_SUCCESS) {
         // Distinguish between "not supported" and "actual error"
         if (IsVideoUnsupportedResult(result)) {
@@ -229,6 +247,10 @@ VkResult EncoderConfigAV1::InitDeviceCapabilities(const VulkanDeviceContext* vkD
         std::cout << "  " << std::left << std::setw(48) << "bidirectionalCompoundReferenceNameMask" << ": 0x" << std::hex << av1EncodeCapabilities.bidirectionalCompoundReferenceNameMask << std::dec << std::endl;
         std::cout << "  " << std::left << std::setw(48) << "maxTemporalLayerCount" << ": " << av1EncodeCapabilities.maxTemporalLayerCount << std::endl;
         std::cout << "  " << std::left << std::setw(48) << "maxSpatialLayerCount" << ": " << av1EncodeCapabilities.maxSpatialLayerCount << std::endl;
+        if (feedback2Requested) {
+            std::cout << "  " << std::left << std::setw(48) << "maxPerPartitionFeedbackEntries" << ": " << videoEncodeFeedback2Capabilities.maxPerPartitionFeedbackEntries << std::endl;
+            std::cout << "  " << std::left << std::setw(48) << "supportedPerPartitionEncodeFeedbackFlags" << ": 0x" << std::hex << videoEncodeFeedback2Capabilities.supportedPerPartitionEncodeFeedbackFlags << std::dec << std::endl;
+        }
     }
 
     result = VulkanVideoCapabilities::GetPhysicalDeviceVideoEncodeQualityLevelProperties<VkVideoEncodeAV1QualityLevelPropertiesKHR, VK_STRUCTURE_TYPE_VIDEO_ENCODE_AV1_QUALITY_LEVEL_PROPERTIES_KHR>
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
@@ -46,6 +46,34 @@ static size_t getFormatTexelSize(VkFormat format)
     }
 }
 
+static const char* getEncodeFeedbackFlagName(VkVideoEncodeFeedbackFlagBitsKHR flag)
+{
+    switch (flag) {
+    case VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR:
+        return "BITSTREAM_BUFFER_OFFSET";
+    case VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR:
+        return "BITSTREAM_BYTES_WRITTEN";
+    case VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_HAS_OVERRIDES_BIT_KHR:
+        return "BITSTREAM_HAS_OVERRIDES";
+    case VK_VIDEO_ENCODE_FEEDBACK_AVERAGE_QUANTIZATION_BIT_KHR:
+        return "AVERAGE_QUANTIZATION";
+    case VK_VIDEO_ENCODE_FEEDBACK_MIN_QUANTIZATION_BIT_KHR:
+        return "MIN_QUANTIZATION";
+    case VK_VIDEO_ENCODE_FEEDBACK_MAX_QUANTIZATION_BIT_KHR:
+        return "MAX_QUANTIZATION";
+    case VK_VIDEO_ENCODE_FEEDBACK_INTRA_PIXELS_BIT_KHR:
+        return "INTRA_PIXELS";
+    case VK_VIDEO_ENCODE_FEEDBACK_INTER_PIXELS_BIT_KHR:
+        return "INTER_PIXELS";
+    case VK_VIDEO_ENCODE_FEEDBACK_SKIPPED_PIXELS_BIT_KHR:
+        return "SKIPPED_PIXELS";
+    case VK_VIDEO_ENCODE_FEEDBACK_PICTURE_PARTITION_COUNT_BIT_KHR:
+        return "PICTURE_PARTITION_COUNT";
+    default:
+        return "UNKNOWN";
+    }
+}
+
 VkResult VkVideoEncoder::CreateVideoEncoder(const VulkanDeviceContext* vkDevCtx,
                                             VkSharedBaseObj<EncoderConfig>& encoderConfig,
                                             VkSharedBaseObj<VkVideoEncoder>& encoder)
@@ -1501,10 +1529,113 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
 
     VkQueryPoolVideoEncodeFeedbackCreateInfoKHR encodeFeedbackCreateInfo =
         {VK_STRUCTURE_TYPE_QUERY_POOL_VIDEO_ENCODE_FEEDBACK_CREATE_INFO_KHR};
+    VkQueryPoolVideoEncodePerPartitionFeedbackCreateInfoKHR perPartitionFeedbackCreateInfo =
+        {VK_STRUCTURE_TYPE_QUERY_POOL_VIDEO_ENCODE_PER_PARTITION_FEEDBACK_CREATE_INFO_KHR};
 
-    encodeFeedbackCreateInfo.pNext = encoderConfig->videoCoreProfile.GetProfile();
-    encodeFeedbackCreateInfo.encodeFeedbackFlags = VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR |
-                                                   VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR;
+    VkVideoEncodeFeedbackFlagsKHR encodeFeedbackFlags =
+        VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR |
+        VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR;
+    const VkVideoEncodeFeedbackFlagsKHR supportedFeedbackFlags =
+        encoderConfig->videoEncodeCapabilities.supportedEncodeFeedbackFlags;
+    const void* feedbackPNext = encoderConfig->videoCoreProfile.GetProfile();
+    bool perPartitionFeedbackEnabled = false;
+
+    const EncoderConfigAV1* av1Config = encoderConfig->GetEncoderConfigAV1();
+    if (av1Config != nullptr) {
+        if (av1Config->enablePictureFeedback) {
+            const VkVideoEncodeFeedbackFlagsKHR qpFeedbackFlags =
+                VK_VIDEO_ENCODE_FEEDBACK_MIN_QUANTIZATION_BIT_KHR |
+                VK_VIDEO_ENCODE_FEEDBACK_MAX_QUANTIZATION_BIT_KHR;
+            const VkVideoEncodeFeedbackFlagsKHR unsupportedQpFlags =
+                qpFeedbackFlags & ~supportedFeedbackFlags;
+            if (unsupportedQpFlags != 0) {
+                std::cout << "Warning: dropping unsupported picture feedback flags: 0x"
+                          << std::hex << unsupportedQpFlags << std::dec << std::endl;
+            }
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_AVERAGE_QUANTIZATION_BIT_KHR;
+            encodeFeedbackFlags |= (supportedFeedbackFlags & qpFeedbackFlags);
+        }
+        if (av1Config->enablePixelCountFeedback) {
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_INTRA_PIXELS_BIT_KHR |
+                                   VK_VIDEO_ENCODE_FEEDBACK_INTER_PIXELS_BIT_KHR;
+        }
+        if (av1Config->enableSkippedPixelCountFeedback) {
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_SKIPPED_PIXELS_BIT_KHR;
+        }
+
+        if (av1Config->enablePerPartitionFeedback) {
+            const VkVideoEncodeFeedback2CapabilitiesKHR& feedback2Caps =
+                encoderConfig->videoEncodeFeedback2Capabilities;
+            if (feedback2Caps.maxPerPartitionFeedbackEntries == 0) {
+                std::cerr << "Per-partition encode feedback is not supported "
+                             "by the implementation for this video profile." << std::endl;
+                return VK_ERROR_FEATURE_NOT_PRESENT;
+            }
+
+            uint32_t maxEntries = std::max(1u, av1Config->maxPerPartitionFeedbackEntries);
+            if (maxEntries > feedback2Caps.maxPerPartitionFeedbackEntries) {
+                std::cout << "Warning: clamping maxPerPartitionFeedbackEntries from " << maxEntries
+                          << " to the supported maximum of "
+                          << feedback2Caps.maxPerPartitionFeedbackEntries << std::endl;
+                maxEntries = feedback2Caps.maxPerPartitionFeedbackEntries;
+            }
+
+            const VkVideoEncodePerPartitionFeedbackFlagsKHR requestedPartitionFlags =
+                VK_VIDEO_ENCODE_PER_PARTITION_FEEDBACK_STATUS_BIT_KHR |
+                VK_VIDEO_ENCODE_PER_PARTITION_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR |
+                VK_VIDEO_ENCODE_PER_PARTITION_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR;
+            const VkVideoEncodePerPartitionFeedbackFlagsKHR partitionFlags =
+                requestedPartitionFlags & feedback2Caps.supportedPerPartitionEncodeFeedbackFlags;
+            if (partitionFlags != requestedPartitionFlags) {
+                std::cout << "Warning: dropping unsupported per-partition feedback flags: 0x"
+                          << std::hex << (requestedPartitionFlags & ~partitionFlags)
+                          << std::dec << std::endl;
+            }
+            if (partitionFlags == 0) {
+                std::cerr << "None of the requested per-partition feedback flags "
+                             "are supported by the implementation." << std::endl;
+                return VK_ERROR_FEATURE_NOT_PRESENT;
+            }
+
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_PICTURE_PARTITION_COUNT_BIT_KHR;
+
+            perPartitionFeedbackCreateInfo.pNext = feedbackPNext;
+            perPartitionFeedbackCreateInfo.maxPerPartitionFeedbackEntries = maxEntries;
+            perPartitionFeedbackCreateInfo.perPartitionEncodeFeedbackFlags = partitionFlags;
+            feedbackPNext = &perPartitionFeedbackCreateInfo;
+            perPartitionFeedbackEnabled = true;
+        }
+    }
+
+    encodeFeedbackCreateInfo.pNext = feedbackPNext;
+    encodeFeedbackCreateInfo.encodeFeedbackFlags = encodeFeedbackFlags;
+
+    const VkVideoEncodeFeedbackFlagsKHR unsupportedFeedbackFlags =
+        encodeFeedbackFlags & ~supportedFeedbackFlags;
+    if (unsupportedFeedbackFlags != 0) {
+        // Report each unsupported flag individually so the user can tell
+        // exactly which requested feedback the implementation lacks.
+        for (uint32_t flagBit = 1; flagBit != 0; flagBit <<= 1) {
+            if ((unsupportedFeedbackFlags & flagBit) == 0) {
+                continue;
+            }
+            std::cerr << "Requested encode feedback '"
+                      << getEncodeFeedbackFlagName(
+                             (VkVideoEncodeFeedbackFlagBitsKHR)flagBit)
+                      << "' (0x" << std::hex << flagBit << std::dec
+                      << ") is not supported by the implementation." << std::endl;
+        }
+        return VK_ERROR_FEATURE_NOT_PRESENT;
+    }
+
+    m_encodeFeedbackFlags = encodeFeedbackFlags;
+    if (perPartitionFeedbackEnabled) {
+        m_perPartitionFeedbackFlags = perPartitionFeedbackCreateInfo.perPartitionEncodeFeedbackFlags;
+        m_maxPerPartitionFeedbackEntries = perPartitionFeedbackCreateInfo.maxPerPartitionFeedbackEntries;
+    } else {
+        m_perPartitionFeedbackFlags = 0;
+        m_maxPerPartitionFeedbackEntries = 0;
+    }
 
     result = m_encodeCommandBufferPool->Configure( m_vkDevCtx,
                                                    encoderConfig->numInputImages, // numPoolNodes
--- a/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
+#include <algorithm>
 #include <chrono>
+#include <cstring>
 #include "VkVideoEncoder/VkVideoEncoderAV1.h"
 #include "VkVideoCore/VulkanVideoCapabilities.h"
 
@@ -84,6 +86,12 @@ VkResult VkVideoEncoderAV1::InitEncoderCodec(VkSharedBaseObj<EncoderConfig>& enc
     VkResult result = InitEncoder(encoderConfig);
     if (result != VK_SUCCESS) {
         fprintf(stderr, "\nERROR: InitEncoder() failed with ret(%d)\n", result);
+        return result;
+    }
+
+    result = m_feedback2Output.Init(*m_encoderConfig);
+    if (result != VK_SUCCESS) {
+        fprintf(stderr, "\nERROR: Feedback2 output init failed with ret(%d)\n", result);
         return result;
     }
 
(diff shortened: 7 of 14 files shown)


elima:
Heads-up for @dabrain34, @lolzballs, @srinathkr-nv

common/include/VkVSCommon.h
```diff
@@ -57,6 +62,59 @@ extern "C" {
 #define VKVS_CASE_STR(x) case x: return VKVS_STRINGIFY(x)
 #endif
 
+// Helper function to get string representation of VkResult codes
+inline const char* string_VkResult(VkResult result) {
```
dabrain34: it appears to be dead code and this can be done using vk_enum_string_helper.h so all the change in VKVSCommon.h can be dropped

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -823,8 +823,17 @@ VkResult VulkanDeviceContext::CreateVulkanDevice(int32_t numDecodeQueues,
                                                                             VK_FALSE
                                                                            };
 
+        VkPhysicalDeviceVideoEncodeFeedback2FeaturesKHR feedback2Features {
```
dabrain34: I would keep the same way as AV1 and VP9 are chained up, using the pNext as it is pretty hard to follow

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -840,6 +849,9 @@ VkResult VulkanDeviceContext::CreateVulkanDevice(int32_t numDecodeQueues,
         CHECK_VULKAN_FEATURE(timelineSemaphoreFeatures.timelineSemaphore, "timelineSemaphore", false);
         CHECK_VULKAN_FEATURE(videoMaintenance1Features.videoMaintenance1, "videoMaintenance1", true);
         CHECK_VULKAN_FEATURE(synchronization2Features.synchronization2, "synchronization2", false);
+        if (m_videoEncodeFeedback2Enabled) {
```
dabrain34: I would use the same as way as AV1 Feature to keep the same syntax

common/libs/VkCodecUtils/VulkanDeviceContext.h
```diff
@@ -111,6 +111,8 @@ class VulkanDeviceContext : public vk::VkInterfaceFunctions {
     bool    GetVideoEncodeQueryResultStatusSupport() const { return m_videoEncodeQueryResultStatusSupport; }
     VkQueueFlags GetVideoDecodeQueueFlag() const { return m_videoDecodeQueueFlags; }
     VkQueueFlags GetVideoEncodeQueueFlag() const { return m_videoEncodeQueueFlags; }
+    void SetVideoEncodeFeedback2Enabled(bool enable) { m_videoEncodeFeedback2Enabled = enable; }
```
dabrain34: could you add  `m_vkDevCtxt.AddReqDeviceExtension(VK_KHR_VIDEO_ENCODE_FEEDBACK_2_EXTENSION_NAME);` inside this method and then remove the if block in vulkan_video_encoder.cpp+137

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
```diff
@@ -850,4 +850,3 @@ bool EncoderConfigH265::InitParamameters(VpsH265 *vpsInfo, SpsH265 *spsInfo,
 
 
 
-
```
dabrain34, line 853: dead line

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
```diff
@@ -1006,9 +1171,134 @@ VkResult VkVideoEncoderAV1::AssembleBitstreamData(VkSharedBaseObj<VkVideoEncodeF
...
+    }
+
+    const char* bitstreamPath = config.outputFileHandler.GetFileName();
+    if (bitstreamPath && bitstreamPath[0] != '\0') {
+        m_path = std::string(bitstreamPath) + ".feedback.txt";
+    } else {
+        m_path = "encoder.feedback.txt";
+    }
+
+    size_t fileSize = m_file.SetFileName(m_path.c_str());
```
dabrain34, line 1206: this call might create a filename which will overflow the maximum of 256 char allocated to m_fileName in VkEncoderConfig.h.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1501,10 +1501,66 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
 
     VkQueryPoolVideoEncodeFeedbackCreateInfoKHR encodeFeedbackCreateInfo =
         {VK_STRUCTURE_TYPE_QUERY_POOL_VIDEO_ENCODE_FEEDBACK_CREATE_INFO_KHR};
+    VkQueryPoolVideoEncodePerPartitionFeedbackCreateInfoKHR perPartitionFeedbackCreateInfo =
+        {VK_STRUCTURE_TYPE_QUERY_POOL_VIDEO_ENCODE_PER_PARTITION_FEEDBACK_CREATE_INFO_KHR};
+
+    VkVideoEncodeFeedbackFlagsKHR encodeFeedbackFlags =
```
dabrain34, line 1535: VkVideoEncodeFeedback2CapabilitiesKHR is never chained into vkGetPhysicalDeviceVideoCapabilitiesKHR anywhere in the repo, so maxPerPartitionFeedbackEntries and supportedPerPartitionEncodeFeedbackFlags are never read from the driver. The code hardcodes all three per-partition flag bits and passes the user's --maxPerPartitionFeedbackEntries through with only a std::max(1u, ...) floor. This violates VUID-VkQueryPoolCreateInfo-queryType-12440 (entries must be ≤ caps) and -12441 (flags must be a subset of supported).

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
```diff
@@ -857,21 +1028,15 @@ VkResult VkVideoEncoderAV1::AssembleBitstreamData(VkSharedBaseObj<VkVideoEncodeF
...
-        uint32_t bitstreamSize;
-        VkQueryResultStatusKHR status;
-    } encodeResult{};
+    EncodeFeedbackResults encodeResult{};
 
     // Fetch the coded VCL data and its information
-    result = m_vkDevCtx->GetQueryPoolResults(*m_vkDevCtx, queryPool, querySlotId,
-                                             1, sizeof(encodeResult), &encodeResult, sizeof(encodeResult),
-                                             VK_QUERY_RESULT_WITH_STATUS_BIT_KHR | VK_QUERY_RESULT_WAIT_BIT);
+    result = GetEncodeFeedbackResults(queryPool, querySlotId, encodeResult);
```
dabrain34, line 1041: The assert are silently passed in Release. Add a case where `encodeResult.status` is tested to different from VK_QUERY_RESULT_STATUS_COMPLETE_KHR and return an error

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1501,10 +1501,66 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
...
+    encodeFeedbackCreateInfo.encodeFeedbackFlags = encodeFeedbackFlags;
+
+    const VkVideoEncodeFeedbackFlagsKHR supportedFeedbackFlags =
+        encoderConfig->videoEncodeCapabilities.supportedEncodeFeedbackFlags;
+    const VkVideoEncodeFeedbackFlagsKHR unsupportedFeedbackFlags =
+        encodeFeedbackFlags & ~supportedFeedbackFlags;
+    if (unsupportedFeedbackFlags != 0) {
+        std::cerr << "Requested encode feedback flags are not supported: 0x"
+                  << std::hex << unsupportedFeedbackFlags << std::dec << std::endl;
+        return VK_ERROR_INITIALIZATION_FAILED;
```
dabrain34: should be VK_ERROR_FEATURE_NOT_PRESENT instead

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1507,15 +1507,18 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
...
     const void* feedbackPNext = encoderConfig->videoCoreProfile.GetProfile();
     bool perPartitionFeedbackEnabled = false;
 
     const EncoderConfigAV1* av1Config = encoderConfig->GetEncoderConfigAV1();
     if (av1Config != nullptr) {
         if (av1Config->enablePictureFeedback) {
-            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_AVERAGE_QUANTIZATION_BIT_KHR |
-                                   VK_VIDEO_ENCODE_FEEDBACK_MIN_QUANTIZATION_BIT_KHR |
-                                   VK_VIDEO_ENCODE_FEEDBACK_MAX_QUANTIZATION_BIT_KHR;
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_AVERAGE_QUANTIZATION_BIT_KHR;
```
lolzballs, line 1555: This commit (`VkVideoEncoder: request only supported qp feedback flags`) should be squashed into `Add support for VK_KHR_video_encode_feedback2 extension for AV1`.

dabrain34:
you should run the tests with `vvs_test_runner.sh` on supported HW

You should not report an issue which is not public

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
elima: Makes sense, thanks.

lolzballs:
Seems like the CTS build from CI is failing because the headers aren't new enough? @dabrain34 are you able to assist with that?

lolzballs:
Oh, the Windows arm64 build is failing because of not new enough headers. The CTS build is failing because it doesn't have the new inls? Not sure what we need to do in this situation, but it seems like it's just an integration issue so maybe we can bypass the CTS check in this case?

elima:
> Oh, the Windows arm64 build is failing because of not new enough headers. The CTS build is failing because it doesn't have the new inls? Not sure what we need to do in this situation, but it seems like it's just an integration issue so maybe we can bypass the CTS check in this case?

That's also what I think we should do. The problem could happen with other extensions too, seems like an issue inherent to this CI job.

We are currently discussing options.

elima:
> you should run the tests with `vvs_test_runner.sh` on supported HW
> 
> You should not report an issue which is not public

Right, my bad. Both have been fixed.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
dabrain34: This should be checked and reported that the requested flags are not supported by the HW otherwise you will silently fails to set it up, see line +1520

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
```diff
@@ -832,12 +840,179 @@ VkResult VkVideoEncoderAV1::SubmitVideoCodingCmds(VkSharedBaseObj<VkVideoEncodeF
...
+    auto addStatus = [&dataSize]() { dataSize += sizeof(VkQueryResultStatusKHR); };
+
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR) {
+        addU32();
+    }
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR) {
+        addU32();
+    }
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_HAS_OVERRIDES_BIT_KHR) {
+        dataSize += sizeof(VkBool32);
```
dabrain34: can you use addU32() instead ? I there a reason to not have a lambda here ? this is only cosmestic comment

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
```diff
@@ -139,6 +139,25 @@ int EncoderConfigAV1::DoParseArguments(int argc, const char* argv[])
                     READ_PARAM(i, lrConfig.LoopRestorationSize[j], uint16_t);
                 }
             }
+        } else if (args[i] == "--pictureFeedback") {
+            enablePictureFeedback = true;
+
+            if (((i + 1) < argc) && (args[i + 1] == "--params")) {
+                ++i;
+                customPictureFeedbackParams = true;
+                READ_PARAM(i, pictureFeedbackAvgQp, int32_t);
```
dabrain34: This is deadcode as pictureFeedbackAvgQp, pictureFeedbackMinQp, pictureFeedbackMaxQp are never used anywhere, please remove also the documentation

dabrain34 (changes requested):

dabrain34:
I think that would be great to have a full unit test in `encode_samples.json` to validate the CLI params.

dabrain34:
I added new tests in `encode_samples.json` 
[encode_samples.json](https://github.com/user-attachments/files/30191039/encode_samples.json)
 to validate the use of the parameters and I hit two different issues:

1.av1_encode_feedback is crashing with 

```
     Warning: dropping unsupported encode feedback flags: 0x100
     vk-video-enc-test: /home/igalia/scerveau/DEV/IGALIA/PROJECTS/VALVE/DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp:1041: virtual VkResult VkVideoEncoderAV1::AssembleBitstreamData(VkSharedBaseObj<VkVideoEncoder::VkVideoEncodeFrameInfo>&, uint32_t, uint32_t): Assertion `encodeResult.status == VK_QUERY_RESULT_STATUS_COMPLETE_KHR' failed.
```

2.  `encode_av1_encode_feedback_per_partition` reports not supported with:

```
   STDERR:
     Per-partition encode feedback is not supported: maxPerPartitionFeedbackEntries 0, unsupported per-partition flags 0x7
     
     ERROR: InitEncoder() failed with ret(-8)
     Can't create the video encoder: unsupported feature
```

on nvidia

On Radv I havent been able to find a working public driver.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -144,6 +144,15 @@ static void printHelp(VkVideoCodecOperationFlagBitsKHR codec)
                                         Eg: 1. \"--cdef --params 3  2  1 2 9 1   2 3 7 3   3 1 4 2   4 0 3 2\"\n\
                                             2. \"--cdef --params 3  0  1 2 9 1\"\n\
                                             3. \"--cdef\"\n\
+        --pictureFeedback               Enable picture-level feedback2 output\n\
+        --params                        Enable custom picture feedback params when followed by --pictureFeedback option\n\
+                                        Otherwise default feedback params will be used\n\
+                                        Following picture feedback parameters must be followed in order with this option\n\
+                                        <avgqp> <minqp> <maxqp>\n\
+        --pixelFeedback                 Enable pixel count feedback2 output (intra/inter)\n\
+        --skippedPixelFeedback          Enable skipped pixel feedback2 output (implies --pixelFeedback)\n\
```
dabrain34: I think the name is a bit misleading with pixelFeedback, can you try to improve the phrasing or the cli param to clarify this use case

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
elima: Good catch. The root cause is the unchecked strcpy() in EncoderOutputFileHandler::SetFileName(); appending .feedback.txt to a bitstream path that already fits m_fileName[256] on its own can push it past the buffer.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
elima: Done. The status is now tested explicitly and returns VK_INCOMPLETE when not STATUS_COMPLETE, mirroring the error handling in VkVideoEncoder::RetrieveData().

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
elima: Fixed. VkVideoEncodeFeedback2CapabilitiesKHR is now chained into the encode capabilities query when feedback2 is requested. The query pool creation fails with VK_ERROR_FEATURE_NOT_PRESENT if per-partition feedback is unsupported, clamps the entry count to the reported maximum, and masks out unsupported per-partition flags.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
elima: Done, changed to VK_ERROR_FEATURE_NOT_PRESENT.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
elima: Done. A warning is now printed listing the min/max QP flags that get dropped when unsupported. Average QP remains requested unconditionally, so it still fails with an error if unsupported.

common/include/VkVSCommon.h
elima: Dropped. The VkVSCommon.h changes are reverted. Nothing actually needed the string helpers, so vk_enum_string_helper.h wasn't required either.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH265.cpp
elima: Oops, reverted. Not sure how that got overlooked. That file is no longer touched by this PR.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigAV1.cpp
elima: Right! Removed, along with the help text.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
elima: Done, using addU32() now, no reason for the odd one out, it was just an oversight.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
elima: Done.

common/libs/VkCodecUtils/VulkanDeviceContext.h
elima: Done. The setter now adds the extension itself and the caller is reduced to a single SetVideoEncodeFeedback2Enabled(enableFeedback2) call.

elima:
> I think that would be great to have a full unit test in `encode_samples.json` to validate the CLI params.

Ok, I will look into it.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
elima: Ok, what about renaming to `--pixelCountFeedback`? Any other suggestion?

elima:
I fixed most of the reported review issues. Pushed one commit for each fix, so that is easier to review. Once the branch is clear, I will squash the fixes into the main (original) commits. Please let me know if you prefer squashing them now.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
elima: Done. feedback2Features is now chained through the same conditional pNext accumulator as AV1/VP9.

elima:
> > I think that would be great to have a full unit test in `encode_samples.json` to validate the CLI params.
> 
> Ok, I will look into it.

Added two AV1 cases to encode_samples.json: one covering picture-level, pixel-count, and skipped-pixel feedback, and another covering per-partition feedback with --maxPerPartitionFeedbackEntries. Unsupported hardware capabilities are reported as N/S by the runner.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
elima: Renamed the options to --pixelCountFeedback and --skippedPixelCountFeedback so they clearly describe aggregate pixel counts rather than per-pixel feedback. Updated the corresponding configuration members, help text, and test invocation as well.

elima:
> I added new tests in `encode_samples.json` [encode_samples.json](https://github.com/user-attachments/files/30191039/encode_samples.json) to validate the use of the parameters and I hit two different issues:
> 
> 1.av1_encode_feedback is crashing with
> 
> ```
>      Warning: dropping unsupported encode feedback flags: 0x100
>      vk-video-enc-test: /home/igalia/scerveau/DEV/IGALIA/PROJECTS/VALVE/DEV/Vulkan-Video-Samples/vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp:1041: virtual VkResult VkVideoEncoderAV1::AssembleBitstreamData(VkSharedBaseObj<VkVideoEncoder::VkVideoEncodeFrameInfo>&, uint32_t, uint32_t): Assertion `encodeResult.status == VK_QUERY_RESULT_STATUS_COMPLETE_KHR' failed.
> ```
> 
>     2. `encode_av1_encode_feedback_per_partition` reports not supported with:
> 
> 
> ```
>    STDERR:
>      Per-partition encode feedback is not supported: maxPerPartitionFeedbackEntries 0, unsupported per-partition flags 0x7
>      
>      ERROR: InitEncoder() failed with ret(-8)
>      Can't create the video encoder: unsupported feature
> ```
> 
> on nvidia
> 
> On Radv I havent been able to find a working public driver.

dabrain34:
When I run on nvidia driver offering `VK_KHR_video_encode_feedback2                 : extension revision 1:

```
$ ./tests/vvs_test_runner.py -t encode_av1_encode_feedback --verbose
```
I get a `not supported` result:

 ```
   STDERR:
     Requested encode feedback flags are not supported: 0x100
     
     ERROR: InitEncoder() failed with ret(-8)
     Can't create the video encoder: unsupported feature
 ```    

Is it expected ?


With 

```
$ ./tests/vvs_test_runner.py -t encode_av1_encode_feedback_per_partition --verbose
```

I get a failure result

```
     Done processing 15 input frames!
     Encoded file's location is at /DEV/Vulkan-Video-Samples/tests/results/test_output_av1_encode_feedback_per_partition.ivf
   STDERR:
     
     encodeResult.status is (0x0) NOT STATUS_COMPLETE! bitstreamStartOffset 0, bitstreamSize 42486
     
     encodeResult.status is (0x0) NOT STATUS_COMPLETE! bitstreamStartOffset 0, bitstreamSize 16344
     
     encodeResult.status is (0x0) NOT STATUS_COMPLETE! bitstreamStartOffset 0, bitstreamSize 17645
     
     encodeResult.status is (0x0) NOT STATUS_COMPLETE! bitstreamStartOffset 0, bitstreamSize 13142
   --- Decoder validation output (last 10 lines) ---
   | status: ERROR, exit code: 1
   | General error -1094995529 at line 204 in file /DEV/Vulkan-Video-Samples/vk_video_decoder/libs/VkDecoderUtils/FFmpegDemuxer.cppNo AVFormatContext provided.Error: Failed to initialize VideoStreamDemuxer for file: /DEV/Vulkan-Video-Samples/tests/results/test_output_av1_encode_feedback_per_partition.ivf
   --- End of decoder validation output ---
```

Can you add more information about your working setup ?

dabrain34:
> > > I think that would be great to have a full unit test in `encode_samples.json` to validate the CLI params.
> > 
> > 
> > Ok, I will look into it.
> 
> Added two AV1 cases to encode_samples.json: one covering picture-level, pixel-count, and skipped-pixel feedback, and another covering per-partition feedback with --maxPerPartitionFeedbackEntries. Unsupported hardware capabilities are reported as N/S by the runner.

This is expected that CI runner is reporting `not supported` as there is no hardware supporting Vulkan Video on these runners.

elima:
> 
> Can you add more information about your working setup ?

I get the exact same results on NVidia (GeForce RTX 4060 Ti / NVidia 595.44.00):

* encode_av1_encode_feedback_per_partition: *Fail*. This is unexpected, This driver should support per-partition feedback.
* encode_av1_encode_feedback: *Not Supported*. This is expected. NVidia only supports per-partition feedback.

On AMD (RADV and a custom, non-public branch of Mesa that supports feedback2), I get:

* encode_av1_encode_feedback_per_partition: *Not Supported*. This is expected, AMD doesn't support per-partition feedback.
* encode_av1_encode_feedback: *Pass*.

I'm looking into the NVidia failure. My preliminary assessment is that the driver is outputting the right feedback2 values, but leaving the final status blank (zero), when it should be status-complete. A similar CTS test passes so this needs further investigation.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
```diff
@@ -832,12 +840,179 @@ VkResult VkVideoEncoderAV1::SubmitVideoCodingCmds(VkSharedBaseObj<VkVideoEncodeF
...
+
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BUFFER_OFFSET_BIT_KHR) {
+        results.hasBitstreamStartOffset = true;
+        results.bitstreamStartOffset = readU32();
+    }
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_BYTES_WRITTEN_BIT_KHR) {
+        results.hasBitstreamSize = true;
+        results.bitstreamSize = readU32();
+    }
+    if (flags & VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_HAS_OVERRIDES_BIT_KHR) {
```
dabrain34, line 950: This flag is not tested by the CLI options. Can you add an option to validate this flag ?

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
```diff
@@ -1501,10 +1501,103 @@ VkResult VkVideoEncoder::InitEncoder(VkSharedBaseObj<EncoderConfig>& encoderConf
...
+                          << std::hex << unsupportedQpFlags << std::dec << std::endl;
+            }
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_AVERAGE_QUANTIZATION_BIT_KHR;
+            encodeFeedbackFlags |= (supportedFeedbackFlags & qpFeedbackFlags);
+        }
+        if (av1Config->enablePixelCountFeedback) {
+            encodeFeedbackFlags |= VK_VIDEO_ENCODE_FEEDBACK_INTRA_PIXELS_BIT_KHR |
+                                   VK_VIDEO_ENCODE_FEEDBACK_INTER_PIXELS_BIT_KHR;
+        }
+        if (av1Config->enableSkippedPixelCountFeedback) {
```
dabrain34, line 1562: I think the user might not know what is supported by the driver so we should validate that the requested flag is supported with a warning in case its not and report `not supported. 
It could be done at the end of the method or for all the flags we are adding (more granular).

dabrain34 (changes requested):

cmake/FindVulkanSDK.cmake
```diff
@@ -108,6 +108,8 @@ MACRO(FIND_VULKAN_SDK minimum_major_version minimum_minor_version minimum_patch_
         # Set Vulkan Loader options to disable tests
         set(BUILD_TESTS OFF CACHE BOOL "Disable Vulkan-Loader tests" FORCE)
 
+        set(VulkanHeaders_VERSION "${minimum_major_version}.${minimum_minor_version}.${minimum_patch_version}")
```
dabrain34: this is unblocking a situation with VulkanLoader build with Win ARM. I would be in favor for a cleaner solution in VulkanLoader but for the moment I would request a comment explaining why this line and how it should be addressed.

elima: Added a comment explaining why VulkanHeaders_VERSION is set here: it unblocks the Vulkan-Loader configure (e.g,  the Windows arm64 build) when the headers come from FetchContent instead of an installed package, plus a TODO to drop it once Vulkan-Loader derives the version itself when Vulkan-Headers is consumed as a subproject.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoder.cpp
elima: Done. Each requested feedback flag is now validated individually against the driver's supportedEncodeFeedbackFlags. Any unsupported flag is reported by name (e.g. SKIPPED_PIXELS) and the encoder fails with VK_ERROR_FEATURE_NOT_PRESENT, which the runner reports as N/S.

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
elima: Added a --hasOverridesFeedback option that requests VK_VIDEO_ENCODE_FEEDBACK_BITSTREAM_HAS_OVERRIDES_BIT_KHR and writes the reported value to the feedback text output (overrides=...). It goes through the same per-flag support validation as the other flags, and is now exercised by the av1_encode_feedback test.

elima:
> 
> I get the exact same results on NVidia (GeForce RTX 4060 Ti / NVidia 595.44.00):
> 
>     * encode_av1_encode_feedback_per_partition: _Fail_. This is unexpected, This driver should support per-partition feedback.
> 
>     * encode_av1_encode_feedback: _Not Supported_. This is expected. NVidia only supports per-partition feedback.
> 
> I'm looking into the NVidia failure. My preliminary assessment is that the driver is outputting the right feedback2 values, but leaving the final status blank (zero), when it should be status-complete. A similar CTS test passes so this needs further investigation.

Regarding the encode_av1_encode_feedback_per_partition failure on NVidia (595.44.00 / RTX 4060 Ti): we believe this is a driver bug rather than an issue in this PR. The encode itself completes normally — all 15 frames are processed and a non-empty bitstream is produced (e.g. bitstreamStartOffset=0, bitstreamSize=42486) — but the whole-picture query status reads back 0x0 (VK_QUERY_RESULT_STATUS_NOT_READY_KHR) on every frame, so the encoder treats the result as incomplete and aborts, leaving an empty .ivf that then fails decode validation. That status is read only after waiting on the encode-complete fence and with VK_QUERY_RESULT_WAIT_BIT set, with vkGetQueryPoolResults returning VK_SUCCESS, so a NOT_READY status is contradictory: the results are finalized and the preceding offset/size fields read back correctly, which indicates the driver populated the feedback values but never wrote the whole-picture status (leaving it at its reset value) instead of VK_QUERY_RESULT_STATUS_COMPLETE_KHR. Notably the AV1 CTS feedback2_partition_count_* tests request per-partition feedback only, whereas this case also enables --pictureFeedback, so the trigger may be the combination of whole-picture QP feedback with per-partition feedback. @srinathkr-nv can you take a look please, in case we are missing something?

srinathkr-nv:
@elima I noticed your above comments just now, after being at-mentioned. Can you try out the latest Vulkan beta driver (Linux 595.44.11, Windows 596.83) from https://developer.nvidia.com/vulkan-driver to see if this issue reproduces with it?

cmake/FindVulkanSDK.cmake
crmurillo: @elima: just fyi, there is a fix now in the Vulkan Loader for this. 
I created #228 because the fix is introduced in **v1.4.358**, but if we update to that version, we need an additional fix for the windows arm64 job.

elima:
> @elima I noticed your above comments just now, after being at-mentioned. Can you try out the latest Vulkan beta driver (Linux 595.44.11, Windows 596.83) from https://developer.nvidia.com/vulkan-driver to see if this issue reproduces with it?

Hi, Srinath!
I could not find 595.44.11, but tried 595.45.04 (nearest I could find) and 610.43.03 (latest) and with both `encode_av1_encode_feedback_per_partition` gives not-supported instead of failure.

I also changed the video sample to one with higher resolution, because the original test had 352x288 which I understand NVidia feedback2 doesn't support. But still the code never reaches that point, the extension is not advertised.

I have spent several rounds of debugging trying to find a culprit on VVS side, but have exhausted all avenues. Is it possible that later drivers like 610.43.03 (and 595.45.04) don't support (or advertise) encode_feedback2?

 ```
$ ./tests/vvs_test_runner.py -t encode_av1_encode_feedback_per_partition --verbose
✓ Found encoder: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_encoder/demos/vk-video-enc-test
✓ Found decoder: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_decoder/demos/vk-video-dec-test
✓ Found executable: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_encoder/demos/vk-video-enc-test
✓ Found executable: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_decoder/demos/vk-video-dec-test
=== Vulkan Video Samples Test Framework ===
Encoder: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_encoder/demos/vk-video-enc-test
Decoder: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_decoder/demos/vk-video-dec-test

✓ All encoder YUV resource files found and verified
✓ All decoder resource files found and verified

==================================================
RUNNING ENCODER TESTS
==================================================
======================================================================
VULKAN VIDEO TEST SUITE
======================================================================
Binary: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_encoder/demos/vk-video-enc-test

✓ All encoder YUV resource files found and verified

[1/1] Running: encode_av1_encode_feedback_per_partition
    Command: /home/devel/devel/src/Vulkan-Video-Samples/build/vk_video_encoder/demos/vk-video-enc-test -i /home/devel/devel/src/Vulkan-Video-Samples/tests/resources/video/yuv/1920x1080_420_8le.yuv --codec av1 --inputWidth 1920 --inputHeight 1080 --inputNumPlanes 3 --verbose --profile main --noDeviceFallback --pictureFeedback --enablePerPartitionFeedback --maxPerPartitionFeedbackEntries 16 -o /home/devel/devel/src/Vulkan-Video-Samples/tests/results/test_output_av1_encode_feedback_per_partition.ivf
✓ Detected driver: nvidia
✓ Detected GPU: NVIDIA GeForce RTX 4060 Ti
  ⚠️  Warning detected in av1_encode_feedback_per_partition
○ N/S (0.47s)
   === Command Output ===
   STDOUT:
     No QP was provided. Using default value: 20.
     numFrames 0 should be different from zero and inferior to input file frame count: 30. Use input file frame count.
     Enumerating instance layers:
     	VK_LAYER_NV_optimus
     	VK_LAYER_NV_present
     	VK_LAYER_VALVE_steam_fossilize_64
     	VK_LAYER_VALVE_steam_overlay_64
     	VK_LAYER_VALVE_steam_fossilize_32
     	VK_LAYER_VALVE_steam_overlay_32
     	VK_LAYER_MESA_device_select
     	VK_LAYER_RENDERDOC_Capture
     	VK_LAYER_MESA_overlay
     	VK_LAYER_KHRONOS_validation
     	VK_LAYER_INTEL_nullhw
     Looking for instance layers:
     Enumerating instance extensions:
     	VK_KHR_device_group_creation
     	VK_KHR_display
     	VK_KHR_external_fence_capabilities
     	VK_KHR_external_memory_capabilities
     	VK_KHR_external_semaphore_capabilities
     	VK_KHR_get_display_properties2
     	VK_KHR_get_physical_device_properties2
     	VK_KHR_get_surface_capabilities2
     	VK_KHR_surface
     	VK_KHR_surface_protected_capabilities
     	VK_KHR_wayland_surface
     	VK_KHR_xcb_surface
     	VK_KHR_xlib_surface
     	VK_EXT_acquire_drm_display
     	VK_EXT_acquire_xlib_display
     	VK_EXT_debug_report
     	VK_EXT_debug_utils
     	VK_EXT_direct_mode_display
     	VK_EXT_display_surface_counter
     	VK_EXT_surface_maintenance1
     	VK_EXT_swapchain_colorspace
     	VK_NV_display_stereo
     	VK_KHR_portability_enumeration
     	VK_LUNARG_direct_driver_loading
     Looking for instance extensions:
     ###### List of Instance Extensions: ######
     	 VK_KHR_device_group_creation(v.1)
     	 VK_KHR_display(v.23)
     	 VK_KHR_external_fence_capabilities(v.1)
     	 VK_KHR_external_memory_capabilities(v.1)
     	 VK_KHR_external_semaphore_capabilities(v.1)
     	 VK_KHR_get_display_properties2(v.1)
     	 VK_KHR_get_physical_device_properties2(v.2)
     	 VK_KHR_get_surface_capabilities2(v.1)
     	 VK_KHR_surface(v.25)
     	 VK_KHR_surface_protected_capabilities(v.1)
     	 VK_KHR_wayland_surface(v.6)
     	 VK_KHR_xcb_surface(v.6)
     	 VK_KHR_xlib_surface(v.6)
     	 VK_EXT_acquire_drm_display(v.1)
     	 VK_EXT_acquire_xlib_display(v.1)
     	 VK_EXT_debug_report(v.10)
     	 VK_EXT_debug_utils(v.2)
     	 VK_EXT_direct_mode_display(v.1)
     	 VK_EXT_display_surface_counter(v.1)
     	 VK_EXT_surface_maintenance1(v.1)
     	 VK_EXT_swapchain_colorspace(v.5)
     	 VK_NV_display_stereo(v.1)
     	 VK_KHR_portability_enumeration(v.1)
     	 VK_LUNARG_direct_driver_loading(v.1)
     *** Selected Vulkan physical device with name: NVIDIA GeForce RTX 4060 Ti, vendor ID: 10de, device UUID: 2939fda2-d1d5-41c2-0ee3-8cfaecad267c, and device ID: 2803, driver ID: 4, driver name: NVIDIA, driver info: 610.43.03, Num Decode Queues: 0, Num Encode Queues: 1 ***
     ================================================================================
                               AV1 Encoder Capabilities
     ================================================================================
       flags                                           : 0xd
       maxSingleReferenceCount                         : 7
       singleReferenceNameMask                         : 0x7f
       maxUnidirectionalCompoundReferenceCount         : 4
       maxUnidirectionalCompoundGroup1ReferenceCount   : 4
       unidirectionalCompoundReferenceNameMask         : 0xf
       maxBidirectionalCompoundReferenceCount          : 7
       maxBidirectionalCompoundGroup1ReferenceCount    : 4
       maxBidirectionalCompoundGroup2ReferenceCount    : 3
       bidirectionalCompoundReferenceNameMask          : 0x7f
       maxTemporalLayerCount                           : 1
       maxSpatialLayerCount                            : 1
       maxPerPartitionFeedbackEntries                  : 0
       supportedPerPartitionEncodeFeedbackFlags        : 0x0
     ================================================================================
                          AV1 Encoder Quality Level Properties
     ================================================================================
       preferredRateControlMode                        : 4
       preferredRateControlLayerCount                  : 1
       preferredRateControlFlags                       : 5
       preferredGopFrameCount                          : 250
       preferredKeyFramePeriod                         : 250
       preferredConsecutiveBipredictiveFrameCount      : 5
       preferredTemporalLayerCount                     : 1
       preferredConstantQIndex.intraQIndex             : 114
       preferredConstantQIndex.predictiveQIndex        : 131
       preferredConstantQIndex.bipredictiveQIndex      : 147
       preferredMaxSingleReferenceCount                : 1
       preferredSingleReferenceNameMask                : 1
       preferredMaxUnidirectionalCompoundReferenceCount: 4
       preferredMaxUnidirectionalCompoundGroup1ReferenceCount: 4
       preferredUnidirectionalCompoundReferenceNameMask: 15
       preferredMaxBidirectionalCompoundReferenceCount : 7
       preferredMaxBidirectionalCompoundGroup1ReferenceCount: 4
       preferredMaxBidirectionalCompoundGroup2ReferenceCount: 3
       preferredBidirectionalCompoundReferenceNameMask : 127
     ================================================================================
                                 Encoder Capabilities
     ================================================================================
       flags                                           : 0x4
       minCodedExtent                                  : 192 x 128
       maxCodedExtent                                  : 8192 x 8192
       pictureAccessGranularity                        : 64 x 64
       encodeInputPictureGranularity                   : 2 x 2
       maxDpbSlots                                     : 17
       maxActiveReferencePictures                      : 7
       maxQualityLevels                                : 7
       maxRateControlLayers                            : 1
       maxBitrate                                      : 120000000
       rateControlModes                                : 0x7
       minBitstreamBufferOffsetAlignment               : 256
       minBitstreamBufferSizeAlignment                 : 256
     
     GOP frame count: 30, IDR period: 30, Consecutive B frames: 5
     
     Input order:   0   1   2   3   4   5   6   7   8   9   10  11  12  13  14  15  16  17  18  19  20  21  22  23  24  25  26  27  28  29  
     Frame Type:   IDR B   B   B   B   B   P   B   B   B   B   B   P   B   B   B   B   B   P   B   B   B   B   B   P   B   B   B   B   P   
     Encode  order: 0   2   3   4   5   6   1   8   9   10  11  12  7   14  15  16  17  18  13  20  21  22  23  24  19  26  27  28  29  25  
     Warning: dropping unsupported picture feedback flags: 0x30
   STDERR:
     Per-partition encode feedback is not supported by the implementation for this video profile.
     
     ERROR: InitEncoder() failed with ret(-8)
     Can't create the video encoder: unsupported feature


======================================================================
VULKAN VIDEO CODEC TEST RESULTS SUMMARY
======================================================================

ENCODER RESULTS:
======================================================================
VULKAN VIDEO ENCODER TEST RESULTS SUMMARY
======================================================================
AV1      -  0 pass,  1 N/S,  0 crash,  0 fail ( 1 total)
----------------------------------------------------------------------
○ av1  encode_av1_encode_feedback_per_partition - N/S   (0.47s)
----------------------------------------------------------------------

### NVIDIA GeForce RTX 4060 Ti / NVIDIA 610.43.03 / Debian GNU/Linux 13 (trixie)

Total Tests:     1
Passed:          0
Crashed:         0
Failed:          0
Not Supported:   1

✔️ ALL TESTS COMPLETED - 0 passed, 1 not supported by hardware/driver

======================================================================
OVERALL SUMMARY
======================================================================

### Debian GNU/Linux 13 (trixie)

Total Tests:     1
Passed:          0
Crashed:         0
Failed:          0
Not Supported:   1

✔️ ALL TESTS COMPLETED - 0 passed, 1 not supported by hardware/driver
📊 Results exported to: /home/devel/devel/src/Vulkan-Video-Samples/tests/results/encode_results.json
🧹 Cleaned up output artifacts
🧹 Cleaned up output artifacts
```

**Comment by srinathkr-nv** [comment]
> Is it possible that later drivers like 610.43.03 (and 595.45.04) don't support (or advertise) encode_feedback2?

Yes, this is expected. `VK_KHR_video_encode_feedback2` is supported only by the NVIDIA Vulkan beta drivers specified in https://developer.nvidia.com/vulkan-driver under the section "Vulkan Developer Beta Driver Release Updates".

The link to the 595.44.11 beta driver is present in the section "Vulkan 1.4 Developer Beta Driver Downloads" but I'll provide it here directly: https://developer.nvidia.com/downloads/assets/gameworks/downloads/secure/Vulkan_Beta_Drivers/NVIDIA-Linux-x86_64-595.44.11.run

**Comment by elima** [comment]
> 
> The link to the 595.44.11 beta driver is present in the section "Vulkan 1.4 Developer Beta Driver Downloads" but I'll provide it here directly: https://developer.nvidia.com/downloads/assets/gameworks/downloads/secure/Vulkan_Beta_Drivers/NVIDIA-Linux-x86_64-595.44.11.run

Thanks, Srinath. I had been testing on 595.45.04 and 610.43.03, which don't advertise the extension, but 595.44.11 does indeed :)

On 595.44.11 (RTX 4060 Ti) per-partition feedback is now advertised and the test reaches the encode. It reproduces the original failure though: the whole-picture query status reads back 0 (VK_QUERY_RESULT_STATUS_NOT_READY_KHR) on every frame, even though the encode completes and writes a valid bitstream (non-zero offset and size).

The status is read with VK_QUERY_RESULT_WAIT_BIT, after also waiting on the encode-complete fence, and vkGetQueryPoolResults returns VK_SUCCESS. So NOT_READY looks wrong here: the result is finalized, and the offset and size fields that precede the status read back correctly, so the whole-picture status field is being left at zero instead of set to COMPLETE. My working assumption is that the per picture partition statuses are correct and only the whole-picture status is affected.

Let me know if you want me to share a dump of the per-partition statuses if that helps.

**Comment by srinathkr-nv** [comment]
I really should have checked the updates to the test application code in this PR earlier. I noticed just now that this PR has updated only the `vk_video_encoder/src/vulkan_video_encoder.cpp` code to call `VulkanDeviceContext::SetVideoEncodeFeedback2Enabled()` but has not made similar updates to the "demo" application at `﻿vk_video_encoder/demos/vk-video-enc/Main.cpp`, which the test runner uses.

Naturally, this means that the encode_feedback2 test cases run without enabling the `videoEncodeFeedback2` device feature, so none of the feedback corresponding to this extension is actually available in the query result buffer. The code in `VkVideoEncoderAV1::GetEncodeFeedbackResults()` expects some of the new feedback values (like the avg, min and max quantization) and reads the result buffer for these values, accidentally consuming the whole-picture status.

**Comment by elima** [comment]
> I really should have checked the updates to the test application code in this PR earlier. I noticed just now that this PR has updated only the `vk_video_encoder/src/vulkan_video_encoder.cpp` code to call `VulkanDeviceContext::SetVideoEncodeFeedback2Enabled()` but has not made similar updates to the "demo" application at `﻿vk_video_encoder/demos/vk-video-enc/Main.cpp`, which the test runner uses.
> 

Great catch! That was indeed the problem; totally overlooked adding the call to the demo.

`per-partition` feedback test is now passing with [3562e09](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/224/commits/3562e093da251121b34a677be19bd0bb0bd8dd6b).

Thank you for the reviews!

File: tests/encode_samples.json [hunk 17]
```diff
@@ -276,6 +276,40 @@
...
     },
+    {
+      "name": "av1_encode_feedback",
+      "codec": "av1",
+      "profile": "main",
+      "extra_args": [
+        "--pictureFeedback",
+        "--pixelCountFeedback",
+        "--skippedPixelCountFeedback",
+        "--hasOverridesFeedback"
```
lolzballs: I would suggest not including `--hasOverridesFeedback` here, it is not an mandatory feedback with feedback2.

dabrain34:
I removed the following commits:

- [build: explain the VulkanHeaders_VERSION workaround in FindVulkanSDK](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/224/commits/7ea259d1cb0f285689e5fe16922d5d95cb6552f7)
[7ea259d](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/224/commits/7ea259d1cb0f285689e5fe16922d5d95cb6552f7)

- [build: fix Vulkan-Loader configure when headers come from FetchContent](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/224/commits/047ebfd210379fc5f1c691fa23701c4cd97a4627)
[047ebfd](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/224/commits/047ebfd210379fc5f1c691fa23701c4cd97a4627)

As now the Windows ARM job is passing since #228 and the following branch has been rebased.

@srinathkr-nv and @lolzballs you can now merged the PR and then I will create a new release version to be used by CTS. Please let me know if you have any requests before doing so.

dabrain34 (approved):

lolzballs:
@dabrain34 thanks for rebasing, but could you squash the fixes like @elima originally proposed?

> I fixed most of the reported review issues. Pushed one commit for each fix, so that is easier to review. Once the branch is clear, I will squash the fixes into the main (original) commits. Please let me know if you prefer squashing them now.

srinathkr-nv:
@dabrain34 since you will be creating a new release, can you go ahead and merge these changes once you squash the commits? It will be faster than pushing a squashed commit, then waiting for me or @lolzballs to merge the changes before tagging a release.

dabrain34:
## Commits squashed

1. 17f88856 ("encoder: Add support for VK_KHR_video_encode_feedback2 extension for AV1", 2026-02-06)

 - 1e02d492 ("VulkanDeviceContext: request feedback2 extension from the setter", 2026-07-20)
 - 9a1ae371 ("VulkanDeviceContext: chain feedback2 features like AV1/VP9", 2026-07-20)
 - 11ec8bcc ("VkVideoEncoderAV1: use addU32() for the has-overrides feedback size", 2026-07-20)
 - 1e188e3e ("VkVideoEncoder: return FEATURE_NOT_PRESENT for unsupported feedback flags", 2026-07-20)
 - 8345e693 ("VkEncoderConfigAV1: remove unused picture feedback QP parameters", 2026-07-20)
 - 8bbea61a ("VkEncoderConfigH265: revert unrelated whitespace change", 2026-07-20)
 - abd4f8bf ("VkVSCommon: drop unused string_VkResult helpers", 2026-07-20)
 - 4a0c6613 ("VkVideoEncoder: return FEATURE_NOT_PRESENT for unsupported feedback flags", 2026-07-20)
 - 56404901 ("encoder: Report output status, bitstream size and offset for feedback2", 2026-02-27)
 - ba45bf0a ("VkVideoEncoderAV1: fail on encode query status not COMPLETE", 2026-07-20)
 - a02c0cca ("encoder: validate per-partition feedback against feedback2 capabilities", 2026-07-20)
 - b59c7308 ("VkVideoEncoder: warn when dropping unsupported QP feedback flags", 2026-07-20)
 - cdcb88ff ("VkVideoEncoder: report unsupported encode feedback flags individually", 2026-07-24)
 - df051321 ("VkVideoEncoder: enable the feedback2 device feature in the demo app", 2026-08-06)
 - 56ea1c7e ("VkEncoderConfigAV1: rename pixel feedback CLI options", 2026-07-21)


2. 99172e5f ("tests: add AV1 encode feedback2 samples", 2026-07-20)

 - e95cbec2 ("tests: use a 1080p sample for the AV1 encode feedback tests", 2026-08-04)
 - 56ea1c7e ("VkEncoderConfigAV1: rename pixel feedback CLI options", 2026-07-21) (partially)

Here is a new history validated with RADV/ANV/nVIdia on Linux.

No change in the code, only fixups for 1 and 2.

Can you please @lolzballs and @srinathkr-nv validate this new history ? and then if you are okay approve the PR. I will then proceed to merge and create the new release.

lolzballs (approved):
Thanks @elima and @dabrain34. LGTM.

srinathkr-nv (approved):

PR #226, opened 2026-07-16
build: fix Windows builds with the new Vulkan Loader

crmurillo:
## Description

Vulkan-Loader configure failure (standalone build)

If the  Vulkan-Loader is updated to v1.4.353, the loader's CMakeLists reads `VulkanHeaders_VERSION` from `find_package(VulkanHeaders)`. When `FIND_VULKAN_SDK` provides the headers via FetchContent there is no installed package to find, the variable stays empty, causing the build to abort with:
 
```if given arguments: "VERSION_GREATER" "1.4.353"```

Only hosts without a system Vulkan SDK (e.g. Windows CI runners) build the loader from source, which is why Linux was unaffected.

## Type of change

bug fix

Total Tests:    83
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%

### DIFF
--- a/cmake/FindVulkanSDK.cmake
+++ b/cmake/FindVulkanSDK.cmake
@@ -108,6 +108,8 @@ MACRO(FIND_VULKAN_SDK minimum_major_version minimum_minor_version minimum_patch_
         # Set Vulkan Loader options to disable tests
         set(BUILD_TESTS OFF CACHE BOOL "Disable Vulkan-Loader tests" FORCE)
 
+        set(VulkanHeaders_VERSION "${minimum_major_version}.${minimum_minor_version}.${minimum_patch_version}")
+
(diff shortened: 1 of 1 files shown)


dabrain34:
I think we are tweaking a potential bug in VulkanLoader itself, so I would be in favor to fix the problem in VulkanLoader and add a comment to tell about this workaround or tweak in VVS waiting for the real fix to happen.

crmurillo:
The issue solved by this PR has been fixed in the Vulkan Loader v1.4.358 by commit [59a705](https://github.com/KhronosGroup/Vulkan-Loader/commit/59a70594de4559fd3aed73fe8ac2d6c48ca002d4).
A new PR will be open to update to the new Vulkan Loader version.

PR #229, opened 2026-08-11
tests: introduce tests comparison

dabrain34:
## Description

In order to check that there is no regression over a new run, introduced a test comparison which will display
if there is a regression or a progression

The file is available in tests/results/vvs_test_results.json and contains a list of tests by hardware_vendor_id_device_id.

This file can be removed and no comparison will be done or a custom file can be passed.

## Type of change

feature

### Intel(R) Graphics (RPL-S) (8086:a780) / Intel open-source Mesa driver Mesa 26.2.0-devel (git-12db716fff) / Ubuntu 24.04.4 LTS

Total Tests:    83
Passed:         57
Crashed:         0
Failed:          0
Not Supported:  23
Success Rate: 100.0%

### NVIDIA GeForce RTX 4060 (10de:2882) / NVIDIA 595.44.00 / Ubuntu 24.04.4 LTS

Total Tests:    83
Passed:         72
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%


tests/vvs_test_runner.py
```diff
@@ -874,6 +896,23 @@ def run_framework_tests(args: argparse.Namespace, encoder_path: str,
     # Print summary
     success = framework.print_summary(encode_results, decode_results)
 
+    # Compare and save results only for full test suite runs (not -t filtered)
+    if not args.no_compare_results and not args.test:
+        system_info = None
+        if framework.decode_framework:
+            system_info = framework.decode_framework.system_info
+        elif framework.encode_framework:
+            system_info = framework.encode_framework.system_info
```
crmurillo, line 907: If I run this command:
`python3 tests/vvs_test_runner.py --encoder-only --save-results`.
the results are not saved because the decoder binary exists but the system_info is empty. 
It results in: `Cannot save results: GPU name not detected`.

tests/vvs_test_runner.py
```diff
@@ -874,6 +896,23 @@ def run_framework_tests(args: argparse.Namespace, encoder_path: str,
     # Print summary
     success = framework.print_summary(encode_results, decode_results)
 
+    # Compare and save results only for full test suite runs (not -t filtered)
+    if not args.no_compare_results and not args.test:
```
crmurillo, line 900: This check is only excluding the "-t" option. If any other option is passed to execute a partial run, like --codec, the results will be saved, but deletes the results of all other tests recorded statuses.

For example, if I execute a full run (encode and decode tests) and then do a partial run of the av1 tests:
`✔️ Summary: 29 unchanged, 54 removed`, the results of the older tests are removed.

tests/libs/video_test_result_storage.py
```diff
@@ -0,0 +1,225 @@
...
+    current: Dict[str, str],
+) -> dict:
+    """Classify test results into regressions, improvements, new, removed."""
+    regressions = []
+    improvements = []
+    new_tests = []
+    unchanged = 0
+    success = VideoTestStatus.SUCCESS.value
+
+    for name, cur_status in sorted(current.items()):
```
crmurillo, line 98: What about other test status changes? for example, from not supported to crash.

tests/libs/video_test_result_storage.py
```diff
@@ -0,0 +1,225 @@
...
+RESULTS_FILENAME = "vvs_test_results.json"
+
+_REGRESSION_MARK = "\u274C"
+_IMPROVEMENT_MARK = "\U0001F389"
+_NO_REGRESSION_MARK = "\u2714\uFE0F"
+
+
+def sanitize_hardware_key(gpu_name: str, vendor_id: str = "",
+                          device_id: str = "") -> str:
+    """Convert GPU name, vendor ID, and device ID to a safe JSON key.
```
crmurillo, line 55: What about the driver version? For example, if I test different mesa versions, am I able to catch regressions between them?

tests/vvs_test_runner.py
```diff
@@ -717,6 +722,23 @@ def create_argument_parser() -> argparse.ArgumentParser:
...
+        help="Do not compare the results to "
+             "vvs_test_results.json keyed by GPU name.")
+    default_results = str(
+        Path(__file__).parent / "results" / RESULTS_FILENAME
+    )
+    parser.add_argument(
+        "--results-file",
+        default=default_results,
+        help="Path to the results JSON file for "
+             "save/compare (default: %(default)s)")
```
crmurillo, line 741: The default file is `$VVS_DIR/tests/results/vvs_test_results.json`.
If the option `--results-file` is passed with the same directory but a different file name: `--results-file tests/results/results.json`, the results file is deleted as part of the cleaning process.

```
Results saved to tests/results/results.json (key: AMD_Radeon_RX_7600_RADV_NAVI33_1002_7480)
📊 Results exported to: $VVS_DIR/tests/results/encode_results.json
🧹 Cleaned up output artifacts
📊 Results exported to: $VVS_DIR/tests/results/decode_results.json
🧹 Cleaned up output artifacts
```
The cleaning step deletes all files in the directory `$VVS_DIR/tests/results/` that do not end with `_results.json`.

tests/libs/video_test_result_storage.py
```diff
@@ -0,0 +1,225 @@
...
+    _print_comparison(diff, hardware_key, previous_timestamp)
+    return len(diff["regressions"]) == 0
+
+
+def save_results(
+    filepath: Path,
+    hardware_key: str,
+    system_info: SystemInfo,
+    results: List[TestResult],
+) -> bool:
```
crmurillo: The function always returns true and the value return value is not checked by the caller.

tests/libs/video_test_result_storage.py
```diff
@@ -0,0 +1,225 @@
...
+    Returns False if regressions are detected, True otherwise.
+    """
+    if not system_info.gpu_name:
+        print("Cannot save results: GPU name not detected")
+        return True
+
+    hardware_key = sanitize_hardware_key(system_info.gpu_name,
+                                         system_info.vendor_id,
+                                         system_info.device_id)
+    filepath = results_file
```
crmurillo: results_file can be used directly without creating another variable.

tests/vvs_test_runner.py
dabrain34: yes I see the pitfall. I'm wondering if we should disable to save the results in the case of --encoder-only or --decoder-only as it will save the partial result.

Anyway this should be fixed now.

tests/libs/video_test_result_storage.py
dabrain34: should be fixed by 302904f

tests/libs/video_test_result_storage.py
dabrain34: thanks for pointing this out ...

tests/libs/video_test_result_storage.py
dabrain34: fixed

tests/vvs_test_runner.py
dabrain34: fixed

tests/vvs_test_runner.py
dabrain34: I think this is not a problem as the original file is tracked by VCS.

tests/libs/video_test_result_storage.py
dabrain34: I would say that this key needs to represent the exact HW but not the driver version as this one is evolving over the time and it is already tracked in the json format as "driver_version": "595.44.00"

tests/vvs_test_runner.py
crmurillo: I'd prefer to disable saving the results in case of a partial run. Normally, I would not upload the results if I only executed a partial run. The current approach also has the disadvantage I mentioned in [discussion_r3790396500](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/229#discussion_r3790396500), where the results file is modified and the tests not executed are removed from the file.

tests/vvs_test_runner.py
crmurillo: Maybe this can be fixed by only saving the results for full runs. The framework should not remove entries from the results file just because the user did not request them.

crmurillo (approved):
LGTM, thanks!

tests/results/vvs_test_results.json
```diff
@@ -0,0 +1,578 @@
+{
+  "NVIDIA_GeForce_RTX_3050_Ti_Laptop_GPU_10de_25a0": {
```
crmurillo: Now that the OS is part of the key, this entry is staled and duplicated with the one in line 98.

tests/skipped_samples.json
```diff
@@ -266,6 +266,51 @@
...
+      "name": "h265_main10_profile",
+      "format": "vvs",
+      "drivers": [
+        "anv"
+      ],
+      "platforms": [
+        "linux"
+      ],
+      "reproduction": "always",
+      "reason": "",
```
crmurillo: The reason and bug_irl can be filled out with the same information as for the test av1_main_profile_10bit.

tests/skipped_samples.json
```diff
@@ -266,6 +266,51 @@
...
+      "format": "vvs",
+      "drivers": [
+        "nvidia"
+      ],
+      "platforms": [
+        "linux"
+      ],
+      "reproduction": "always",
+      "reason": "",
+      "bug_url": "",
```
crmurillo: bug_url can be filled with https://github.com/KhronosGroup/Vulkan-Video-Samples/issues/231, in case it is the same error.

tests/skipped_samples.json
```diff
@@ -266,6 +266,51 @@
       "reason": "",
       "bug_url": "",
       "date_added": "2026-01-20"
+    },
```
crmurillo, line 269: The tests are failing for ANV, as reported by the test comparison. 
They can be added to the skipped list.
- encode_av1_encode_feedback
- encode_av1_encode_feedback_per_partition

tests/manage_samples_list.py
```diff
@@ -413,31 +427,31 @@ def remove_skip_entries(skip_list: Dict) -> None:
...
-                      f"{', '.join(entry['drivers'])})")
+        device = get_input("Enter device ID (blank for any)",
+                           allow_empty=True)
+
+        for section, entries in (('decode', decode_entries),
+                                 ('encode', encode_entries)):
+            for i, entry in enumerate(entries):
+                if driver not in entry.get('drivers', []):
+                    continue
+                if device and device not in entry.get('devices', []):
```
crmurillo, line 448: Is a normalized value required here? Maybe _device_matches() can be reused.

tests/manage_samples_list.py
```diff
@@ -57,6 +58,10 @@
 VALID_PLATFORMS = ["all", "linux", "windows"]
 VALID_REPRODUCTION = ["always", "flaky"]
 
+# Skip-list device IDs: VkPhysicalDeviceProperties deviceID as hex digits,
+# with optional fnmatch wildcards
+_DEVICE_ID_RE = re.compile(r'[0-9a-fA-F*?\[\]!-]+')
```
crmurillo, line 63: The device pattern is different than what is in _device_matches() in video_test_config_base.py.
I assume they should be the same, maybe use a common pattern function.

dabrain34: yes because the input can be anything. It could be shared but `manage_samples_list.py` is pretty independent from test libs

tests/manage_samples_list.py
dabrain34: it can be improved indeed but yes it has to be normalized as this is a random input

PR #230, opened 2026-08-11
tests: probe hardware via vulkaninfo before running tests

dabrain34:
## Description

Use vulkaninfo to detect supported video extensions and the GPU driver before executing tests. This avoids crashes by skipping unsupported codec tests early and introduces a category field ("none", "flaky", "hard_crash") in the skip list so that known crash-inducing tests are never launched when the matching driver is confirmed.

If vulkaninfo is not available, all checks are silently skipped and behaviour is unchanged.

## Type of change

refactor

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.2.0-devel (git-12db716fff) / Ubuntu 24.04.4 LTS

Total Tests:    83
Passed:         69
Crashed:         0
Failed:          0
Not Supported:  10
Success Rate: 100.0%

### DIFF
--- a/tests/libs/video_test_framework_base.py
+++ b/tests/libs/video_test_framework_base.py
@@ -28,6 +28,7 @@ from typing import Dict, List, Optional
 
 from tests.libs.video_test_config_base import (
     BaseTestConfig,
+    SkipCategory,
     SkipFilter,
     SkipRule,
     TestResult,
@@ -50,6 +51,9 @@ from tests.libs.video_test_utils import DEFAULT_TEST_TIMEOUT
 # Exit codes from sysexits.h
 EX_OK = 0                 # Successful termination
 EX_UNAVAILABLE = 69       # Service unavailable (used for "not supported")
+
+# A dry run only initializes the device; it must never take as long as a test.
+DRY_RUN_TIMEOUT = 60
 
 # Unix signals (as positive return codes from wait())
 SIGABRT = 6               # Abort signal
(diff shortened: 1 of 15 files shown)


dabrain34:
@lolzballs can you give it look ?

tests/libs/video_test_framework_base.py
```diff
@@ -119,6 +123,51 @@ def __init__(self, executable_path: str = None, **options):
...
+            self._probed_driver = None
+        if self._hw_extensions is not None and self.verbose:
+            print(f"✓ Hardware probe via vulkaninfo succeeded"
+                  f" (driver: {self._probed_driver or 'unknown'})")
+        if gpu_count > 1:
+            driver = self._probed_driver or "unknown"
+            print(f"WARNING: {gpu_count} GPUs detected, probe "
+                  f"uses first device ({driver}). May be "
+                  f"inaccurate with --deviceID.")
+            if input("Continue with hardware probe? [y/N] "
```
crmurillo, line 142: Double prompt for "Continue with hardware probe?". The probe is called in the __init__ function of each framework: once for encode and once for decode, causing the double prompt, even if only a decode test is executed. For example,
`python3 tests/vvs_test_runner.py -t decode_av1_argon_test787`. 
The probe could be called somewhere else where it is only executed once.

tests/libs/video_test_framework_base.py
```diff
@@ -119,6 +123,51 @@ def __init__(self, executable_path: str = None, **options):
         self._system_info: Optional[SystemInfo] = None
         self._test_pattern_active = False
         self._skipped_samples: Dict[str, Optional[SkipRule]] = {}
+        hw_ext, probed_driver, gpu_count = probe_vulkan_video_support()
+        self._hw_extensions = hw_ext
+        if probed_driver:
+            self._probed_driver = DriverMapping.normalize_driver_name(
```
crmurillo, line 129: Hard crashes for ANV are not skipped. The driverName printed for Mesa 26.2.0 is "Intel open-source Mesa driver", there is no mapping for such string in DRIVER_NAME_MAPPING.

tests/libs/video_test_framework_base.py
```diff
@@ -119,6 +123,51 @@ def __init__(self, executable_path: str = None, **options):
         self._system_info: Optional[SystemInfo] = None
         self._test_pattern_active = False
         self._skipped_samples: Dict[str, Optional[SkipRule]] = {}
+        hw_ext, probed_driver, gpu_count = probe_vulkan_video_support()
```
crmurillo, line 126: The probe is currently enabled by default, it is executed even when passing the option `--ignore-skip-list`.
Does it make sense to have the hardware probe disabled by default and a command line option to enable it? For normal every day testing, it does not seem necessary to have the probe running for every test, specially when running single tests with the `-t` option.

crmurillo: There is also the use case when working on fixing a crash.
To be able to verify a fix for a test that was crashing, it is now necessary to pass the option `--ignore-skip-tests` to execute it, otherwise, the test is always skipped.

dabrain34:
Will adopt  a dryRun approach

PR #236, opened 2026-08-26
tests: handle hard_crash category in skipped_samples.json

dabrain34:
## Description

Add a way to dry run encode and decode process to detect the hardware before running the test, avoiding to crash the hardware such as with some tests.

So a dryRun detects the hardware and just skip the test to run only if the category is hard_crash.

## Type of change

feature

### AMD Radeon RX 7600 (RADV NAVI33) / radv Mesa 26.1.0-devel (git-16ece32984) / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         70
Crashed:         0
Failed:          0
Not Supported:  11
Success Rate: 100.0%

### DIFF
--- a/common/libs/VkCodecUtils/DecoderConfig.h
+++ b/common/libs/VkCodecUtils/DecoderConfig.h
@@ -60,6 +60,7 @@ struct DecoderConfig {
         validateVerbose = false;
 
         noPresent = false;
+        dryRun = false;
 
         maxFrameCount = -1;
         videoFileName = "";
@@ -187,6 +188,15 @@ struct DecoderConfig {
                 "Runs this program headless without presenting decode result to "
                 "screen",
                 [this](const char **args, const ProgramArgs &a) {
+                    noPresent = true;
+                    return true;
+                }},
+            {"--dryRun", nullptr, 0,
+                "Probe the hardware only: decode up to the first sequence header so the "
+                "video session and the decode capabilities get validated, then exit "
+                "without decoding any frame (implies --noPresent)",
+                [this](const char **args, const ProgramArgs &a) {
+                    dryRun = true;
                     noPresent = true;
                     return true;
                 }},
@@ -461,6 +471,7 @@ struct DecoderConfig {
     uint32_t validateVerbose : 1;
     uint32_t verbose : 1;
     uint32_t noPresent : 1;
+    uint32_t dryRun : 1;
     uint32_t enableHwLoadBalancing : 1;
     uint32_t selectVideoWithComputeQueue : 1;
     uint32_t outputy4m : 1;
--- a/tests/libs/video_test_dry_run_probe.py
+++ b/tests/libs/video_test_dry_run_probe.py
@@ -1,0 +1,49 @@
+"""
+Video Test Dry Run Probe
+Hardware capability probing via the applications' --dryRun mode.
+
+Copyright 2026 Igalia S.L.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import subprocess
+from pathlib import Path
+from typing import Optional, Tuple
+
+from tests.libs.video_test_platform_utils import PlatformUtils
+
+# A dry run only initializes the device; it must never take as long as a test.
+DRY_RUN_TIMEOUT = 60
+
+
+def run_dry_run(cmd: Optional[list], *, unsupported_exit_code: int,
+                cwd: Optional[Path] = None) -> Tuple[bool, str, str]:
+    """Run a dry-run command; report support plus the output it produced.
+
+    The support verdict is True whenever it cannot be determined, so that a
+    test is never skipped on the strength of a probe that failed to run.
+    """
+    if not cmd:
+        return True, "", ""
+
+    try:
+        subprocess_kwargs = PlatformUtils.get_subprocess_kwargs()
+        subprocess_kwargs['timeout'] = DRY_RUN_TIMEOUT
+        if cwd is not None:
+            subprocess_kwargs['cwd'] = str(cwd)
+        result = subprocess.run(cmd, check=False, **subprocess_kwargs)
+    except (OSError, subprocess.SubprocessError):
+        return True, "", ""
+
+    return (result.returncode != unsupported_exit_code), result.stdout, result.stderr
--- a/tests/libs/video_test_framework_base.py
+++ b/tests/libs/video_test_framework_base.py
@@ -28,6 +28,7 @@ from typing import Dict, List, Optional
 
 from tests.libs.video_test_config_base import (
     BaseTestConfig,
+    SkipCategory,
     SkipFilter,
     SkipRule,
     TestResult,
@@ -38,13 +39,15 @@ from tests.libs.video_test_config_base import (
 )
 
 from tests.libs.video_test_platform_utils import PlatformUtils
+from tests.libs.video_test_dry_run_probe import run_dry_run
 from tests.libs.video_test_driver_detect import (
     parse_driver_from_output, parse_system_info_from_output, SystemInfo,
     get_os_info
 )
 from tests.libs.video_test_result_reporter import (
-    get_status_display, print_codec_breakdown, print_detailed_results,
-    print_final_summary, print_command_output)
+    count_results_by_status, get_status_display, group_results_by_codec,
+    print_codec_breakdown, print_detailed_results, print_final_summary,
+    print_command_output, result_to_dict)
 from tests.libs.video_test_utils import DEFAULT_TEST_TIMEOUT
 
 # Exit codes from sysexits.h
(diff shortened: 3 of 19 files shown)


common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -682,20 +682,19 @@ VkResult VulkanDeviceContext::InitPhysicalDevice(int32_t deviceId, const vk::Dev
...
-                              << ", device UUID: " << currentDeviceUuid.ToString()
-                              << ", and device ID: " << devProp2.properties.deviceID << std::dec
-                              << ", driver ID: " << deviceDriverProperties.driverID
-                              << ", driver name: " << deviceDriverProperties.driverName
-                              << ", driver info: " << deviceDriverProperties.driverInfo
-                              << ", Num Decode Queues: " << m_videoDecodeNumQueues
-                              << ", Num Encode Queues: " << m_videoEncodeNumQueues
-                              << " ***" << std::endl << std::flush;
-                }
+
```
crmurillo: Remove the extra empty line here. Also line 686.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
```diff
@@ -682,20 +682,19 @@ VkResult VulkanDeviceContext::InitPhysicalDevice(int32_t deviceId, const vk::Dev
...
+
+                vk::DeviceUuidUtils currentDeviceUuid(deviceVulkan11Properties.deviceUUID);
```
crmurillo: Is it necessary to print this unconditionally?

tests/libs/video_test_framework_base.py
```diff
@@ -759,6 +810,47 @@ def _should_mask_as_skipped(
...
+        if config.name in self._skipped_samples:
+            skip_rule = self._skipped_samples[config.name]
+            reason = skip_rule.reason if skip_rule else "In skip list"
+            return early(f"Skipping: {test_name} ({reason})",
+                         VideoTestStatus.SKIPPED, f"Skipped: {reason}")
+
+        # Checked before the dry-run probe: the probe launches the very
+        # binary a hard_crash rule exists to keep from running.
+        hard_crash_rule = self._is_hard_crash_skip(config.name, test_type)
+        if hard_crash_rule is not None:
```
crmurillo, line 798: A test marked as hard_crash is always skipped, even when you want to execute only that test by passing the `-t` option. 
E.g. `tests/vvs_test_runner.py -t decode_av1_argon_test787` skips the test due to the dryRun, making it difficult to debug the issue unless you pass `--ignore-skip-list`. 
Running individual tests should always execute them, same as it is currently done if the test is marked as skipped.

common/libs/VkCodecUtils/DecoderConfig.h
```diff
@@ -190,6 +191,15 @@ struct DecoderConfig {
                     noPresent = true;
                     return true;
                 }},
+            {"--dryRun", nullptr, 0,
+                "Probe the hardware only: decode up to the first sequence header so the "
+                "video session and the decode capabilities get validated, then exit "
+                "without decoding any frame (implies --noPresent)",
```
crmurillo, line 197: It seems that it actually starts decoding. The check that stops the run lives in the app and only executes after the parser finishes a chunk, and by then the first frames have already been submitted to the GPU. On the argon sample 787 two pictures are decoded before the "no frame decoded" message prints.

tests/libs/video_test_framework_base.py
```diff
@@ -759,6 +810,47 @@ def _should_mask_as_skipped(
...
+            reason = hard_crash_rule.reason or "hard crash"
+            driver = self._detected_driver
+            note = f"known to hard crash on {driver}: {reason}"
+            return early(f"Not run: {test_name} ({note})",
+                         VideoTestStatus.SKIPPED, f"Not run, {note}")
+
+        # Hardware codec support (--dryRun probe)
+        if not self._is_hw_supported(config):
```
crmurillo: This probe is executed unconditionally once per test, what is the added benefit of that?
Unsupported cases are already detected by a normal run.

Also, the output is not descriptive of the actual reason why the test is not supported and can be misleading
```
[29/31] Not supported: encode_h264_intrarefresh (no HW encode support for h264)
[30/31] Not supported: encode_h265_intrarefresh (no HW encode support for h265)
[31/31] Not supported: encode_av1_intrarefresh (no HW encode support for av1)
```

tests/libs/video_test_framework_base.py
```diff
@@ -120,6 +122,52 @@ def __init__(self, executable_path: str = None, **options):
...
         self._skipped_samples: Dict[str, Optional[SkipRule]] = {}
 
+    def build_dry_run_command(self, config) -> Optional[list]:
+        """Command initializing the app without processing a frame - to be
+        implemented by subclasses
+
+        Returns None when no command can be built, which makes the hardware
+        probe a no-op for that test.
+        """
+        raise NotImplementedError(
```
crmurillo: The docstring says returning None makes the probe a no-op, and both callers already handle None. Suggest returning None here, like _default_run_cwd does, and dropping the two MockFramework overrides added in the unit tests.

tests/unit_tests/test_status_determination.py
```diff
@@ -42,6 +42,10 @@ def create_test_suite(self):
         """Create test suite - mock implementation returns empty list."""
         return []
 
+    def build_dry_run_command(self, _config):
```
crmurillo: Can be removed. See [comment](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/236/changes/0cc42b8e064b6f7b0985f98614205d9db20d853b#diff-71f7ce48a0ba039fc7e3efa2658b3ce452e50ec45301959bb571d4d23b209aa2R132).

tests/unit_tests/test_filter_suite.py
```diff
@@ -74,6 +74,10 @@ def create_test_suite(self):
         """Create test suite - mock implementation returns empty list."""
         return []
 
+    def build_dry_run_command(self, _config):
```
crmurillo: Can be removed. See [comment](https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/236/changes/0cc42b8e064b6f7b0985f98614205d9db20d853b#diff-71f7ce48a0ba039fc7e3efa2658b3ce452e50ec45301959bb571d4d23b209aa2R132).

tests/libs/video_test_dry_run_probe.py
```diff
@@ -0,0 +1,53 @@
...
+        subprocess_kwargs['timeout'] = DRY_RUN_TIMEOUT
+        if cwd is not None:
+            subprocess_kwargs['cwd'] = str(cwd)
+        result = subprocess.run(cmd, check=False, **subprocess_kwargs)
+    except (OSError, subprocess.SubprocessError):
+        return True, "", ""
+
+    # A negative returncode means the probe died on a signal (SIGSEGV etc.),
+    # so it determined nothing.
+    supported = (result.returncode < 0
```
crmurillo: returncode < 0 is redundant here, since a negative code already satisfies != unsupported_exit_code. unsupported_exit_code is 69.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -596,7 +600,14 @@ int EncoderConfig::ParseArguments(int argc, const char *argv[])
         return -1;
     }
 
-    if (!outputFileHandler.HasFileName()) {
+    if (!outputFileName.empty() && !dryRun) {
```
crmurillo: These two blocks (lines 603 and 610) only differ in where the name comes from. 
The code can be refactored, assign the default into outputFileName when it is empty, then do a single SetFileName under one !dryRun check.

vk_video_decoder/libs/VkVideoDecoder/VkVideoDecoder.h
```diff
@@ -175,6 +175,10 @@ class VkVideoDecoder : public IVulkanVideoDecoderHandler {
 
     VkResult GetLastResult() const { return m_lastVkResult; }
 
+    // True once StartVideoSequence() has validated the stream against the device
```
crmurillo: Use the same format for function comments as in the rest of the file.

common/libs/VkCodecUtils/VulkanDeviceContext.cpp
dabrain34: you are right that used by CTS it might generate additional verbose. I will bring back the verbose gate

tests/libs/video_test_framework_base.py
dabrain34: or you can remove the hard_skip in the list to debug it. I'm afraid of hard crash because the machine needs to be reboot, that's why I created this section, so I would be in favor to user action in the skip llist or passing --ignore-skip-list because with `-t decode_av1*` you would hit the issue without willing

tests/libs/video_test_framework_base.py
dabrain34: yes you are right, fixing.

common/libs/VkCodecUtils/DecoderConfig.h
dabrain34: DecodePictureWithParameters was indeed called by the picture parser for the case of AV1 and H265.
Now the method early exits in case of dryRun

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.cpp
```diff
@@ -197,10 +198,9 @@ int EncoderConfig::ParseArguments(int argc, const char *argv[])
             if (!vk::IsValidFilePath(args[i].c_str(), false)) {
                 return -1;
             }
-            size_t fileSize = outputFileHandler.SetFileName(args[i].c_str());
-            if (fileSize <= 0) {
-                return (int)fileSize;
-            }
+            // Opening is deferred until the whole command line is parsed, so that
+            // --dryRun given after -o still avoids creating the file.
```
crmurillo, line 202: If I execute the following example, the dry.264 file is still created, even though '--dryRun' was passed as an argument after '-o'. 
```
./build/vk_video_encoder/demos/vk-video-enc-test -i tests/resources/video/yuv/352x288_15_i420.yuv --inputWidth 352 --inputHeight 288 -c avc --deviceID 0x7480 --noDeviceFallback -o dry.264 --dryRun
```

dabrain34: this is due to the change to clean the use of `!dryRun`. https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/236#discussion_r3925348117 

For me this is minor I would be in favor to reintegrate the change

crmurillo: ok, sounds good to me as well to reintegrate the change.

crmurillo (approved):

PR #232, opened 2026-08-19
tests: do not fail encode tests the decoder cannot validate

crmurillo:
## Description

Decoder validation is executed after every successful encode and any non-zero decoder exit marked the encode test as failed. That mixes two different scenarios: an encoded file that is actually broken, and a decoder that has no support for the encoded configuration.

Print a warning in case the decoder does not have the capabilities to validate the encoded file and mark the encode test as successful. The decoder reports this before decoding anything, so there is no information about the validity of the encoded file.

See #231 for details and how to reproduce.

## Type of change

bug fix

## Issue (optional)

Fixes part of #231 .

### NVIDIA GeForce RTX 4060 Ti / NVIDIA 595.44.00 / Ubuntu 24.04.4 LTS

Total Tests:    83
Passed:         72
Crashed:         0
Failed:          0
Not Supported:  10
Skipped:         1 (in skip list)
Success Rate: 100.0%

PR #238, opened 2026-09-01
VkEncoderConfigH264: fix use of B-frame count in profile autoselect logic

srinathkr-nv:
## Description

The H.264 encoder profile auto-selection logic currently uses the value returned by VkVideoGopStructure::GetConsecutiveBFrameCount() for determining the lower and upper bounds on the codec profile which can be used for encoding the input content. However, this can end up selecting a higher profile than strictly necessary because GetConsecutiveBFrameCount() returns a default value of 255 (i.e. CONSECUTIVE_B_FRAME_COUNT_MAX_VALUE) when the user has not explicitly specified B-frames in the command line.

This change adds an extra check for this condition (in which case the B-frame count is set to 0 for use by the auto-selection logic).

## Type of change

Bug fix

### NVIDIA L30 (10de:26bb) / NVIDIA 595.44.15 / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         72 (1 warning(s))
Crashed:         0
Failed:          0
Not Supported:  11
Skipped:         2 (in skip list)
Success Rate: 100.0%

Follow-up to https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/171

### DIFF
--- a/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
+++ b/vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfigH264.cpp
@@ -371,6 +371,10 @@ StdVideoH264ProfileIdc EncoderConfigH264::GetNextLowerProfile(StdVideoH264Profil
 
 VkResult EncoderConfigH264::InitVideoProfileCapabilities(const VulkanDeviceContext* vkDevCtx)
 {
+    const uint8_t consecutiveBFrameCount = gopStructure.GetConsecutiveBFrameCount();
+    const uint8_t inputBFrames =
+        (consecutiveBFrameCount == CONSECUTIVE_B_FRAME_COUNT_MAX_VALUE) ? 0 : consecutiveBFrameCount;
+
     // Track whether the profile was user-specified or auto-selected.
     const bool autoSelected = (profileIdc == STD_VIDEO_H264_PROFILE_IDC_INVALID);
 
@@ -383,7 +387,7 @@ VkResult EncoderConfigH264::InitVideoProfileCapabilities(const VulkanDeviceConte
         profileIdc = STD_VIDEO_H264_PROFILE_IDC_BASELINE;
 
         // Upgrade to MAIN profile if using B-frames or CABAC entropy coding
-        if ((gopStructure.GetConsecutiveBFrameCount() > 0) || (entropyCodingMode == ENTROPY_CODING_MODE_CABAC))
+        if ((inputBFrames > 0) || (entropyCodingMode == ENTROPY_CODING_MODE_CABAC))
             profileIdc = STD_VIDEO_H264_PROFILE_IDC_MAIN;
 
         if (use8x8Transform) {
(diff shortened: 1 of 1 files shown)


dabrain34:
thank you very much for this PR

dabrain34:
I see one crash and one warning in the tests section, could you check what are the failing tests ?

dabrain34 (approved):

srinathkr-nv:
> I see one crash and one warning in the tests section, could you check what are the failing tests ?

The crashing test is `decode_av1_superres_8bit`. This one has always failed for me as far as I can remember.

The test with a warning is `encode_h264_high444_profile` because the output cannot be validated by decoding (NVIDIA's H.264 decoder does not support 4:4:4). This one also used to fail earlier but passes with a warning now due to https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/232

dabrain34:
```
AV1      -  1 pass,  0 N/S,  0 crash,  0 fail ( 1 total)
----------------------------------------------------------------------
✔️ av1  decode_av1_superres_8bit            - PASS  (3.78s)
----------------------------------------------------------------------

### NVIDIA GeForce RTX 4060 (10de:2882) / NVIDIA 595.44.14 / Ubuntu 24.04.4 LTS

Total Tests:     1
```
I checked a normal playback of `tests/resources/video/av1/av1-1920x1080-main-superres-8.ivf` and no issue so far with `NVIDIA GeForce RTX 4060 (10de:2882)`

I dont have the issue neither with `NVIDIA_GeForce_RTX_3050_Ti_Laptop_GPU_10de_25a0_Linux`

I would recommend to investigate why this is failing with `NVIDIA L30`. Can you open an issue ?


Regarding `encode_h264_high444_profile` this is indeed expected.

srinathkr-nv:
I tried out the `decode_av1_superres_8bit` test on an NVIDIA RTX A4000 (Ampere, same GPU family as the the GeForce RTX 3050) and I don't see any failure there.

The GeForce RTX 4060 is of the same GPU family (Ada) as the NVIDIA L30, so it's not clear to me why the test passes on the 4060 and fails on the L30. I'll try to locate a 4060 (or equivalent) and independently verify the test results there. At this point, this is seeming like a bug specific to an NVIDIA GPU family or architecture and I will file a bug report internally at NVIDIA for tracking. @dabrain34 do you also need a GitHub issue opened for this?

dabrain34:
If its identified as a driver bug I would say this is not necessary. you can create a skipped sample entry in tests/skipped_samples,json and add the device id (25a0 for example ) which should be displayed in the testing framework such as 

```
### NVIDIA GeForce RTX 3050 Ti Laptop GPU (10de:25a0) / NVIDIA 595.44.14 / Ubuntu 24.04.4 LTS

Total Tests:    85
Passed:         66 (1 warning(s))
Crashed:         0
Failed:          0
Not Supported:  18
Success Rate: 100.0%
```
Can you rebase your branch to main ?

Can you also show the the crash log with:

`./tests/vvs_test_runner.py -t decode_av1_superres_8bit -v`

Then when its okay you can update the PR description to avoid having a crash in the tests section

srinathkr-nv:
I have created https://github.com/KhronosGroup/Vulkan-Video-Samples/pull/241 to add an entry in skipped_samples.json for the crashing test. As I wrote in that PR, the crash is due to fence timeout and subsequent assert in debug builds.

Once that PR is merged, I'll rebase this one on top of that, re-run the test framework and update the test scores in the PR description here.

dabrain34:
Thanks #241  has been merged, you can rebase and update the description :)

srinathkr-nv:
> Thanks #241 has been merged, you can rebase and update the description :)

Great; thank you for promptly merging that change. I have updated the test results for the current PR.

PR #244, opened 2026-09-15
Helpers: remove an useless warning when vkWaitForFences() timeout

hakzsam:


dabrain34:
can you explain why this is useless? We detected issues with some drivers related to this. I would be in favor to keep this log by now.

hakzsam:
It's spamming the CI job output a lot. If you want to keep it, can we increase the timeout or only print the warning with an envvar or so?

dabrain34:
I'm sorry about that.
How do you use VVS in a CI? Asking as you might be interested to use the testing framework in tests folder.

What driver is spamming a lot on your side ? 

we can put this log under  a verbose condition using --verbose and error out when the number of attempts have been reached.

hakzsam:
Here's an example https://gitlab.freedesktop.org/hakzsam/mesa/-/jobs/110297202

It's just running VKCTS via deqp-runner on Mesa/RADV in Mesa CI.

Making less verbose would be cool yeah

hakzsam:
If you explain how you want this to be done, I can probably do it.

lolzballs:
> can you explain why this is useless? We detected issues with some drivers related to this. I would be in favor to keep this log by now.

I don't really see why the warning would help? The error below would trigger if any real issue happened.

dabrain34:
IIUC the warning has been kept to track a timeout symptom, I agree that this is not always a fatal failure and so the 200ms timeout log could be remove to keep only the 5000ms but it also demonstrates a strange behavior as this is not expected to have a timeout here AFAIK. I noticed that with some drivers it happens from time to time. That's why I recommend a silent mode and not remove the log.

@krajunv ^

hakzsam:
When a bunch of video tests run in parallel, it's not surprising to have some timeout I think.

dabrain34:
There are other use case where the log happens.

I'm reviving #5 to address the issue

hakzsam:
Any progress? Can we make sure this doesn't stall again? It's not like urgent but it would be nice to address it. Thanks!

dabrain34:
Dear @hakzsam,

Yes #5 is now ready. I left a week for people to review it and then without any hand raised I will merge it.

hakzsam:
Great, thanks a lot!

PR #245, opened 2026-09-18
VkEncoderConfig: report file mapping failure as out of memory

souravpNV:
## Description

The file handlers returned an error code from a function whose return type is size_t, so callers treated the value as a byte count. Since size_t is unsigned, the <= 0 guard at each call site only ever caught zero, and zero in turn means success to CreateCodecConfig(). An input file that could not be mapped was therefore ignored: encoder creation reported success and the encode failed much later, dereferencing a null pointer while loading the first frame.

Return int64_t from SetFileName() and OpenFile() so that a negative error propagates, and report a mapping failure as an out-of-memory condition so that callers can distinguish a resource limit from a file that is simply unreadable. Mapping a multi-gigabyte input file cannot succeed in a 32-bit process, and the error reported by the platform in that case varies with load, so the condition is identified by the mapping step failing rather than by the specific error code.

## Type of change

 bug fix

### NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition (10de:2bb4) / NVIDIA 630.11 / Windows 11

Total Tests:    85
Passed:         73 (1 warning(s))
Crashed:         0
Failed:          0
Not Supported:  11
Success Rate: 100.0%

✔️ ALL TESTS COMPLETED - 73 passed, 11 not supported by hardware/driver


vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
dabrain34, line 510: you should return an error here.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -529,12 +536,12 @@ class EncoderOutputFileHandler
...
 
 private:
-    size_t OpenFile()
+    int64_t OpenFile()
     {
         m_fileHandle = fopen(m_fileName, "wb");
         if (m_fileHandle == nullptr) {
             fprintf(stderr, "Failed to open output file %s", m_fileName);
-            return 0;
+            return -1;
```
dabrain34, line 544: maybe you should use also a descriptive enum here such as FILE_OPEN_ERROR as it is inconsistent with OUT_OF_MEMORY_ERROR

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -627,12 +634,12 @@ class EncoderQpMapFileHandler
...
 
 private:
-    size_t OpenFile()
+    int64_t OpenFile()
     {
         m_fileHandle = fopen(m_fileName, "rb");
         if (m_fileHandle == nullptr) {
             fprintf(stderr, "Failed to open input file %s", m_fileName);
-            return 0;
+            return -1;
```
dabrain34, line 642: the same here

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -641,14 +648,16 @@ class EncoderQpMapFileHandler
             fprintf(stderr, "Failed to map the input file %s", m_fileName);
             const auto& errmsg = error.message();
             std::printf("error mapping file: %s, exiting...\n", errmsg.c_str());
-            return error.value();
+            // The file exists and opened, so a mapping failure here can only mean the address
```
dabrain34, line 651: The file can be zero length too so I would comment that it could be the case and should be a different error.

dabrain34 (changes requested):

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: Done

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: Done

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: Thanks for the review, done as suggested.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -529,12 +555,12 @@ class EncoderOutputFileHandler
...
+    int64_t OpenFile()
     {
         m_fileHandle = fopen(m_fileName, "wb");
         if (m_fileHandle == nullptr) {
             fprintf(stderr, "Failed to open output file %s", m_fileName);
-            return 0;
+            return FILE_OPEN_ERROR;
         }
 
         return 1;
```
dabrain34: I think it should return 0 here as this is the real size of the file when we open it

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
```diff
@@ -1205,8 +1205,8 @@ VkResult VkVideoEncoderAV1::Feedback2TextOutput::Init(const EncoderConfigAV1& co
         m_path = "encoder.feedback.txt";
     }
 
-    size_t fileSize = m_file.SetFileName(m_path.c_str());
-    if (fileSize == 0) {
+    int64_t fileSize = m_file.SetFileName(m_path.c_str());
```
dabrain34: fileSize is a bit misleading for SetFileName result but seen that its opening the file after this is correct.
The test should be `<` only after while

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -627,12 +659,23 @@ class EncoderQpMapFileHandler
...
+    int64_t OpenFile()
     {
         m_fileHandle = fopen(m_fileName, "rb");
         if (m_fileHandle == nullptr) {
             fprintf(stderr, "Failed to open input file %s", m_fileName);
-            return 0;
+            return FILE_OPEN_ERROR;
+        }
+
+        // An empty file cannot be mapped either, so rule it out here to keep it distinct
```
dabrain34: All this block could be tracked by the mmap failure errors such as std::errc::invalid_argument or std::errc::not_enough_memory.
I'm not in favor of using fseek and ftell.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -421,12 +435,21 @@ class EncoderInputFileHandler
...
+    int64_t OpenFile()
     {
         m_fileHandle = fopen(m_fileName, "rb");
         if (m_fileHandle == nullptr) {
             fprintf(stderr, "Failed to open input file %s", m_fileName);
-            return 0;
+            return FILE_OPEN_ERROR;
+        }
+
+        // A zero length file cannot be mapped either, and is not an out-of-memory condition.
```
dabrain34: All this block could be tracked by the mmap failure errors such as std::errc::invalid_argument or std::errc::not_enough_memory.
I'm not in favor of using fseek and ftell.

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
dabrain34: This m_memMapedFile is never used so please remove it

dabrain34 (changes requested):

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: OK, I dropped fseek/ftell and used std::filesystem::file_size().

I did try classifying from the error condition first, but on Windows it cannot separate the two cases. A file too large to map reports ERROR_INVALID_PARAMETER (raw 87) and an empty file reports ERROR_INVALID_HANDLE (raw 6), and MSVC maps both to invalid_argument

vk_video_encoder/libs/VkVideoEncoder/VkVideoEncoderAV1.cpp
souravpNV: agreed, done

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: Removed, along with the sync()/unmap() in Destroy() and the GetFileSize() accessor, since both only read it. The output file is written through m_fileHandle, so nothing else used the mapping

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
souravpNV: done

vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h
```diff
@@ -627,28 +646,38 @@ class EncoderQpMapFileHandler
...
         m_memMapedFile.map(m_fileName, 0, mio::map_entire_file, error);
         if (error) {
             fprintf(stderr, "Failed to map the input file %s", m_fileName);
             const auto& errmsg = error.message();
             std::printf("error mapping file: %s, exiting...\n", errmsg.c_str());
-            return error.value();
+
+            // The mapping error does not distinguish a zero length file from running out
+            // of address space, so use the size to tell them apart.
+            std::error_code sizeError;
```
dabrain34:
```suggestion
            ```
           if (error == std::errc::invalid_argument) {
               return FILE_EMPTY_ERROR;
            }
            if (error == std::errc::not_enough_memory) {
                return OUT_OF_MEMORY_ERROR;
            }
            return FILE_OPEN_ERROR;
```
```

Comment (souravpNV) [reply vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h]: On Windows a file too large to map returns ERROR_INVALID_PARAMETER (raw 87), which maps to invalid_argument too, so a large clip would be reported as empty.

It probably works on Linux, where mmap gives EINVAL for empty and ENOMEM for too large, Windows maps both to EINVAL. Happy to go either way if you prefer the error-code version.

Comment (dabrain34) [reply vk_video_encoder/libs/VkVideoEncoder/VkEncoderConfig.h]: Lets keep the check for emptiness and use 

```
            if (error == std::errc::not_enough_memory) {
                return OUT_OF_MEMORY_ERROR;
            }
            return FILE_OPEN_ERROR;
```

souravpNV: done

dabrain34:
As you can see I changed the enum to a dedicated enum to VKVS that I moved to VkVSCommon.h. This is a fixup as you can review it and approve it if you are okay with this change. If its good for you, we will merge this change.

